PR #119, opened 2023-11-13
Edit systems #18

### LINKED ISSUE #18: As an operator/admin, I want to be able to edit systems
- [ ] Implement Logic for Editing Systems from Database
      - Should they be permanently deleted?
- [ ] Implement System Edit Endpoint Class
- [ ] Implement API Endpoint for Editing Systems

joelvdavies:
## Description
Adds a PATCH endpoint for modifying systems.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes #18

### DIFF
--- a/test/e2e/test_system.py
+++ b/test/e2e/test_system.py
@@ -2,6 +2,7 @@
 End-to-End tests for the System router
 """
 
+from typing import Any, Optional
 from unittest.mock import ANY
 
 from bson import ObjectId
@@ -103,6 +104,29 @@ def _post_n_systems(test_client, number):
     )
 
 
+def _test_partial_update_system(
+    test_client, update_values: dict[str, Any], additional_expected_values: Optional[dict[str, Any]] = None
+):
+    """
+    Utility method that tests updating a system
+
+    :param update_values: Values to update
+    :param additional_expected_values: Any additional values expected from the output e.g. code
+    """
+    if additional_expected_values is None:
+        additional_expected_values = {}
+
+    # Create one to update
+    response = test_client.post("/v1/systems", json=SYSTEM_POST_A)
+    system = response.json()
+
+    # Update
+    response = test_client.patch(f"/v1/systems/{system['id']}", json=update_values)
+
+    assert response.status_code == 200
+    assert response.json() == {**system, **update_values, **additional_expected_values}
+
+
 def test_create_system(test_client):
     """
     Test creating a System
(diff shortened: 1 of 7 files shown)


test/e2e/test_system.py
```diff
@@ -417,3 +384,174 @@ def test_get_system_breadcrumbs_with_non_existent_id(test_client):
...
+
+
+def test_partial_update_system_importance(test_client):
+    """
+    Test updating a System's importance
+    """
+    _test_partial_update_system(test_client, {"importance": "high"})
+
+
+def test_partial_update_system_invalid(test_client):
```
VKTB, line 483:
```suggestion
def test_partial_update_system_invalid_id(test_client):
```

test/e2e/test_system.py
```diff
@@ -417,3 +384,174 @@ def test_get_system_breadcrumbs_with_non_existent_id(test_client):
...
+    """
+    Test updating a System when the ID is invalid
+    """
+    response = test_client.patch("/v1/systems/invalid", json={"name": "Updated name"})
+
+    assert response.status_code == 404
+    assert response.json()["detail"] == "A System with such ID was not found"
+
+
+def test_partial_update_system_non_existent(test_client):
```
VKTB, line 493:
```suggestion
def test_partial_update_system_non_existent_id(test_client):
```

test/e2e/test_system.py
```diff
@@ -417,3 +384,174 @@ def test_get_system_breadcrumbs_with_non_existent_id(test_client):
...
+    Test updating a System's owner
+    """
+    _test_partial_update_system(test_client, {"owner": "Updated owner"})
+
+
+def test_partial_update_system_importance(test_client):
+    """
+    Test updating a System's importance
+    """
+    _test_partial_update_system(test_client, {"importance": "high"})
```
VKTB, line 480: We probably don't need these because we don't have any logic that changes them or uses them for something else so we just trust that the dict merging etc works as expected. No harm in keeping them in but this was the only reason why I didn't test fields like this in the catalogue category or catalogue item entities. It gets long when the entity has many fields.

VKTB (commented):
Looks good, just some very minor comments.

test/e2e/test_system.py
joelvdavies: Yeah I agree, although if I don't do this I would rather have one that just does all the fields at once instead, unlikely we would mess up the schema but if we did it would at least catch it via that and the others that only update name/parent id.

VKTB (approved):

PR #124, opened 2023-11-27
Add items

### LINKED ISSUE #11: As an operator/admin, I want to be able to create items
- [ ] Implement Item Database Entity Class
      - The entity class should reference a catalogue item
      - The entity class should reference a system
      - The entity class should reference/include order info
      - The entity class should reference a supplier (This means that a Supplier entity class will need to be implemented too)
      - Common fields (and their types) across items need to be defined
      - Which common fields are mandatory vs optional needs to be defined
      - The fields specified at the catalogue category level (in addition to the common fields) which are present in the catalogue should too be present and populated (if needed) when an item is added. By default , the value of the fields should be the same as those in the catalogue item but it should be possible for these values to be changed at an item level. If a value is changed at an item level then the value should not be updated in the catalogue item. This way, the values in the catalogue items are the nominal ones and the values in the item are the actual ones.
- [ ] Implement Logic for Creating Items in Database
- [ ] Implement Item Service Domain Class
- [ ] Implement Item Create Endpoint Classes
- [ ] Implement API Endpoint for Creating Items
- [ ] Address catalogue item TODOs
- [ ] Address item TODOs

VKTB:
## Description
This PR creates an items endpoint (accessible at `/v1/items`) to which `POST` requests can be sent to create new items. The endpoint allows for the catalogue item properties to be overridden at this level through the `catalogue_item_override_properties` field. This does not override the actual catalogue item properties in the corresponding catalogue item itself but keeps them separate. Each catalogue item overrides property object requires a `name`, which is the name of the property as defined in the category, and a `value` for it.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Test creating an item
- [ ] Test creating an item without catalogue item override properties
- [ ] Test creating an item with an invalid catalogue item ID
- [ ] Test creating an item with a non-existent catalogue item ID
- [ ] Test creating an item with an invalid catalogue category ID in a catalogue item
- [ ] Test creating an item with a non-existent catalogue category ID in a catalogue item
- [ ] Test creating an item with an invalid system item ID
- [ ] Test creating an item with a non-existent system item ID
- [ ] Test creating an item with invalid value types for the catalogue item override properties

## Agile board tracking
closes #11

### DIFF
--- a/inventory_management_system_api/services/utils.py
+++ b/inventory_management_system_api/services/utils.py
@@ -4,6 +4,16 @@ Collection of some utility functions used by services
 
 import logging
 import re
+from numbers import Number
+from typing import Dict, Union, List
+
+from inventory_management_system_api.core.exceptions import (
+    MissingMandatoryCatalogueItemProperty,
+    InvalidCatalogueItemPropertyTypeError,
+)
+from inventory_management_system_api.models.catalogue_category import CatalogueItemProperty
+from inventory_management_system_api.schemas.catalogue_category import CatalogueItemPropertyType
+from inventory_management_system_api.schemas.catalogue_item import PropertyPostRequestSchema
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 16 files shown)


joelvdavies (commented):
Minor comments from what I looked at, had 3 files left but they will probably change anyway.

test/e2e/test_item.py
```diff
@@ -0,0 +1,246 @@
...
+ITEM_POST_EXPECTED = {
+    **ITEM_POST,
+    "id": ANY,
+    "purchase_order_number": None,
+    "asset_number": None,
+    "catalogue_item_override_properties": [{"name": "Property A", "value": 21, "unit": "mm"}],
+}
+
+
+def test_create_item(test_client):
```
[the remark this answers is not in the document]
VKTB: Is what I have done in `test/e2e/test_item.py` as part of https://github.com/ral-facilities/inventory-management-system-api/pull/124/commits/d0aa7f5d605a00241412e6e08b4ba9760a80c87b enough?

inventory_management_system_api/services/utils.py
```diff
@@ -23,3 +33,152 @@ def generate_code(name: str, entity_type: str) -> str:
...
     return re.sub(r"\s+", "-", name)
+
+
+def process_catalogue_item_properties(
+    defined_properties: List[CatalogueItemProperty],
+    supplied_properties: List[PropertyPostRequestSchema],
+) -> List[Dict]:
+    """
+    Process and validate supplied catalogue item properties based on defined catalogue item properties. It checks
+    for missing mandatory catalogue item properties unless otherwise instructed, filters the matching catalogue item
```
joelvdavies, line 44:
```suggestion
    for missing mandatory catalogue item properties, filters the matching catalogue item
```

inventory_management_system_api/services/utils.py
```diff
@@ -23,3 +33,152 @@ def generate_code(name: str, entity_type: str) -> str:
...
+    # Convert properties to dictionaries for easier lookups
+    defined_properties_dict = _create_catalogue_item_properties_dict(defined_properties)
+    supplied_properties_dict = _create_catalogue_item_properties_dict(supplied_properties)
+
+    # Some mandatory catalogue item properties may not have been supplied
+    _check_missing_mandatory_catalogue_item_properties(defined_properties_dict, supplied_properties_dict)
+    # Catalogue item properties that have not been defined may have been supplied
+    supplied_properties_dict = _filter_matching_catalogue_item_properties(
+        defined_properties_dict, supplied_properties_dict
+    )
```
joelvdavies, line 62: Is there a reason you chose this over a 422 or something if more properties are supplied? The only reason I can see a problem is if an optional property has a wrong spelling or something it will just ignore it rather than complain.

joelvdavies (commented):
Just tried testing it, I did find giving 
  "catalogue_item_override_properties": []
doesn't return any properties when it should have populated them with the default.

inventory_management_system_api/services/utils.py
VKTB: It mimics the behaviour of Pydantic. If you provide extra fields to Pydantic that do not match any of the defined fields, it will just ignore them.

joelvdavies (approved):
LGTM

PR #133, opened 2023-12-13
Duplicate catalogue item properties #123

### LINKED ISSUE #123: Duplicate catalogue item properties get ignored when creating a catalogue item
Given the below catalogue category:

```
{
  "name": "test",
  "is_leaf": true,
  "parent_id": "655ca5731c251a2a828ca907",
  "catalogue_item_properties": [
    {
      "name": "test",
      "type": "number",
      "unit": "",
      "mandatory": false
    },
    {
      "name": "test",
      "type": "string",
      "unit": "",
      "mandatory": false
    }
  ],
  "id": "655dcd2141fe51dd7b13ea28",
  "code": "test"
}
```

When I create the below catalogue item:

```

{
  "catalogue_category_id": "655dcd2141fe51dd7b13ea28",
  "name": "edsdda",
  "cost_gbp": 3213,
  "cost_to_rework_gbp": null,
  "days_to_replace": 3213,
  "days_to_rework": null,
  "description": null,
  "item_model_number": null,
  "is_obsolete": false,
  "obsolete_reason": null,
  "obsolete_replacement_catalogue_item_id": null,
  "drawing_link": null,
  "drawing_number": null,
  "properties": [
    { "name": "test", "value": 12 },
    { "name": "test", "value": "ewqewq" }
  ],
  "manufacturer": {
    "name": "eseqw",
    "address": "ewqeqe",
    "url": "https://example.com"
  }
}
```

I get the response below which only includes one of the properties that I stated in the post request object above: 

```
{
  "catalogue_category_id": "655dcd2141fe51dd7b13ea28",
  "manufacturer": {
    "name": "eseqw",
    "url": "https://example.com/",
    "address": "ewqeqe"
  },
  "name": "edsdda",
  "description": null,
  "cost_gbp": 3213.0,
  "cost_to_rework_gbp": null,
  "days_to_replace": 3213.0,
  "days_to_rework": null,
  "drawing_number": null,
  "drawing_link": null,
  "item_model_number": null,
  "is_obsolete": false,
  "obsolete_reason": null,
  "obsolete_replacement_catalogue_item_id": null,
  "properties": [{ "name": "test", "value": "ewqewq", "unit": "" }],
  "id": "655dd42d41fe51dd7b13ea2b"
}
```

MatteoGuarnaccia5:
## Description
Added logic to the creation and updating of catalogue categories to make sure that duplicate catalogue item property names are not possible. The logic also ensures that not case specific names (e.g. Property A and property a) are also seen as duplicates

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] check creating a catalogue category with duplicate catalogue item property names
- [ ] check editing a catalogue category to have duplicate catalogue item property names

## Agile board tracking

closes #123

### DIFF
--- a/inventory_management_system_api/services/catalogue_category.py
+++ b/inventory_management_system_api/services/catalogue_category.py
@@ -60,7 +60,7 @@ class CatalogueCategoryService:
 
 
         logger.info(catalogue_category.catalogue_item_properties)
-        if self.check_duplicate_property_names(catalogue_category.catalogue_item_properties if catalogue_category.catalogue_item_properties else []):
+        if self.check_duplicate_property_names(catalogue_category.catalogue_item_properties if catalogue_category.catalogue_item_properties else [], 'post'):
             raise DuplicatePropertyName("Cannot add catalogue category with duplicate catalogue item property names")
 
         code = utils.generate_code(catalogue_category.name, "catalogue category")
@@ -125,6 +125,7 @@ class CatalogueCategoryService:
         update_data = catalogue_category.model_dump(exclude_unset=True)
 
         stored_catalogue_category = self.get(catalogue_category_id)
+
         if not stored_catalogue_category:
             raise MissingRecordError(f"No catalogue category found with ID: {catalogue_category_id}")
 
@@ -137,6 +138,9 @@ class CatalogueCategoryService:
             if parent_catalogue_category and parent_catalogue_category.is_leaf:
                 raise LeafCategoryError("Cannot add catalogue category to a leaf parent catalogue category")
 
+        if "catalogue_item_properties" in update_data and catalogue_category.catalogue_item_properties != stored_catalogue_category.catalogue_item_properties:
+            if self.check_duplicate_property_names(update_data["catalogue_item_properties"], 'patch'):
+                raise DuplicatePropertyName("Cannot edit a catalogue category to have duplicate catalogue item property names")
         # If any of these, need to ensure the category has no children
         if any(key in update_data for key in CATALOGUE_CATEGORY_WITH_CHILDREN_NON_EDITABLE_FIELDS):
             if self._catalogue_category_repository.has_child_elements(CustomObjectId(catalogue_category_id)):
@@ -149,7 +153,7 @@ class CatalogueCategoryService:
         )
 
     def check_duplicate_property_names(self,
-        properties: List[CatalogueItemPropertySchema]
+        properties: List[CatalogueItemPropertySchema], request: str
 ) -> bool:
         """
         Go through all the properties to check for any duplicate property names
@@ -159,9 +163,19 @@ class CatalogueCategoryService:
         """
 
         logger.info('Checking for duplicate property names')
+
         list_of_names = []
-        for dictionary in properties:
-            list_of_names.append(dictionary.name.lower())
+
+        if request == 'post':
+            for dictionary in properties:
+                list_of_names.append(dictionary.name.lower())
+
+        if request == 'patch':
+
+            for dict in properties:
+                for key, value in dict.items():
+                    if key == 'name':
+                        list_of_names.append(value.lower())
         
         duplicate_names = set()
         unique_names = set()
--- a/inventory_management_system_api/services/catalogue_category.py
+++ b/inventory_management_system_api/services/catalogue_category.py
@@ -58,10 +58,9 @@ class CatalogueCategoryService:
         if parent_catalogue_category and parent_catalogue_category.is_leaf:
             raise LeafCategoryError("Cannot add catalogue category to a leaf parent catalogue category")
 
-        if self.check_duplicate_property_names(
+        self.check_duplicate_property_names(
             catalogue_category.catalogue_item_properties if catalogue_category.catalogue_item_properties else [], "post"
-        ):
-            raise DuplicatePropertyName("Cannot add catalogue category with duplicate catalogue item property names")
+        )
 
         code = utils.generate_code(catalogue_category.name, "catalogue category")
         return self._catalogue_category_repository.create(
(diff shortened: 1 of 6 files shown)


inventory_management_system_api/services/catalogue_category.py
```diff
@@ -140,3 +153,36 @@ def update(
...
+                        list_of_names.append(value)
+
+        unique_names = set()
+
+        for name in list_of_names:
+            if name.lower() in unique_names:
+                raise DuplicatePropertyName(
+                    f"Cannot have duplicate catalogue item property name: {name}"
+                )
+            unique_names.add(name.lower())
```
VKTB, line 188: A few things:
1. Lets refer to the properties as `catalogue_item_properties` at this level and keep it consistent with the rest of the module.
2. The method should be private as it's only accessed in this class so prefix it with `_`.
3. We no longer need the `request` arg if you pass `catalogue_category.catalogue_item_properties` to this method in the `update` method.
4. The method no longer returns so the type hint needs to be updated.
5. The docstring needs to be updated accordingly and also include `:raises ` info.
6. I don't think we need `list_of_names` as we can just loop through the `catalogue_item_properties`

```suggestion
    def _check_duplicate_catalogue_item_property_names(
        self, catalogue_item_properties: List[CatalogueItemPropertySchema]
    ) -> None:
        """
        Go through all the catalogue item properties to check for any duplicate names.

        :param catalogue_item_properties: The supplied catalogue item properties
        :raises DuplicateCatalogueItemPropertyName: If a duplicate catalogue item property name is found.
        """
        logger.info("Checking for duplicate catalogue item property names")
        seen_catalogue_item_property_names = set()
        for catalogue_item_property in catalogue_item_properties:
            catalogue_item_property_name = catalogue_item_property.name

            if catalogue_item_property_name.lower() in seen_catalogue_item_property_names:
                raise DuplicateCatalogueItemPropertyNameError(
                    f"Duplicate catalogue item property name: {catalogue_item_property_name}"
                )
            seen_catalogue_item_property_names.add(catalogue_item_property_name.lower())
```

test/e2e/test_catalogue_category.py
```diff
@@ -1013,3 +1031,23 @@ def test_partial_update_catalogue_category_non_existent_id(test_client):
...
     assert response.status_code == 404
     assert response.json()["detail"] == "A catalogue category with such ID was not found"
+
+
+def test_partial_update_catalogue_items_to_have_duplicate_property_names(test_client):
+    """
+    Test updating a catalogue category to have duplicate catalogue item property names
+    """
+    response = test_client.post("/v1/catalogue-categories", json=CATALOGUE_CATEGORY_POST_C)
+    catalogue_category__id = response.json()["id"]
```
VKTB, line 1041:
```suggestion
    catalogue_category_id = response.json()["id"]
```

test/unit/services/test_catalogue_category.py
```diff
@@ -217,6 +219,62 @@ def test_create_with_leaf_parent_catalogue_category(
...
+        CatalogueItemPropertySchema(name="Property A", type="boolean", mandatory=True),
+    ]
+
+    with pytest.raises(DuplicatePropertyName) as exc:
+        # pylint: disable=duplicate-code
+        catalogue_category_service.create(
+            CatalogueCategoryPostRequestSchema(
+                name=catalogue_category.name,
+                is_leaf=catalogue_category.is_leaf,
+                catalogue_item_properties=catalogue_category_catalogue_item_properties,
```
VKTB, line 269: You can do what you did in `test_update_properties_to_have_duplicate_names` and get rid of the `catalogue_category_catalogue_item_properties` var that way
```suggestion
                catalogue_item_properties=[prop.model_dump() for prop in catalogue_category.catalogue_item_properties],
```

test/unit/services/test_catalogue_category.py
```diff
@@ -217,6 +219,62 @@ def test_create_with_leaf_parent_catalogue_category(
...
+        CatalogueCategoryOut(
+            id=catalogue_category.parent_id,
+            name="Category A",
+            code="category-a",
+            is_leaf=False,
+            parent_id=None,
+            catalogue_item_properties=[],
+        ),
+    )
+    # pylint: enable=duplicate-code
```
VKTB, line 255: I don't think a parent catalogue category is needed for the purposes of this test?

VKTB (changes requested):

VKTB:
Could you please also merge `develop` into this branch as it is not up to date?

VKTB (approved):
LGTM. It would be nice if @joelvdavies has quick a look at it before merging it.

test/unit/services/test_catalogue_category.py
```diff
@@ -664,9 +698,6 @@ def test_update_change_catalogue_item_properties_when_has_children(
     )
     # pylint: enable=duplicate-code
 
-    # Mock so child elements found
-    catalogue_category_repository_mock.has_child_elements.return_value = True
-
```
joelvdavies, line 669: Why was this removed? It looks like it should be here based on the name, but then again the tests are passing.

joelvdavies (commented):
Just one thing.

test/unit/services/test_catalogue_category.py
MatteoGuarnaccia5: to be honest I can't remember why, it probably was on accident, but the test passes with and without it so up to you if I should add it back in

joelvdavies: I think it should be as the others typically have them - my guess is that the count_documents function is mocked and so returns an MagicMock or something that is being evaluated to true, which works but I think its clearer if its there.

inventory_management_system_api/services/catalogue_category.py
```diff
@@ -140,3 +149,21 @@ def update(
...
+    def _check_duplicate_catalogue_item_property_names(
+        self, catalogue_item_properties: List[CatalogueItemPropertySchema]
+    ) -> None:
+        """
+        Go through all the catalogue item properties to check for any duplicate names.
+        :param catalogue_item_properties: The supplied catalogue item properties
+        :raises DuplicateCatalogueItemPropertyName: If a duplicate catalogue item property name is found.
+        """
+        logger.info("Checking for duplicate catalogue item property names")
+        seen_catalogue_item_property_names = set()
```
joshuadkitenge, line 162: The property names needs to be trimmed before being checked. Currently if you have "name" and "name " (with a space at the end) they are two different values.


  
dup item prop name

joshuadkitenge (changes requested):

joshuadkitenge (approved):

Issue #128, opened 2023-12-06
Allow catalogue items to be moved between categories only if the catalogue item properties are fully matching


VKTB:
At the moment, catalogue items can be moved from one catalogue category to another as long as the properties of the catalogue item match the list of mandatory properties (and their types) of the catalogue category that it is being moved to. It ignores any supplied properties not defined in the catalogue category and does not take units into account which could be an issue. The moving is done through the `PATCH` `/{catalogue_item_id}` catalogue item endpoint by supplying the `catalogue_category_id` that the catalogue item is being moved to.

Instead, I think that we should do this:
1. Get the current catalogue category of the catalogue item
2. Get the catalogue category that the catalogue item is being moved to
3. Compare the catalogue item property objects from 1 and 2 and allow for the catalogue item to be moved ONLY if they fully match. It looks like Pydantic allows for objects to be compared so this should work with how the `CatalogueItemProperty` model is defined at the moment but if not we should look at overriding the model's `__eq__` method. Please note that the order of the catalogue item property objects may differ so we may need to order them in some way before doing the comparison.

Ref https://github.com/ral-facilities/inventory-management-system-api/issues/9

VKTB:
The same should be done for moving items.

PR #152, opened 2024-01-15
edit item #13

### LINKED ISSUE #13: As an end-operator/admin, I want to be able to edit items
- [ ] Implement Logic for Editing Items in Database
- [ ] Implement Item Edit Endpoint Class
- [ ] Implement API Endpoint for Editing Items

MatteoGuarnaccia5:
## Description
Implemented api endpoint to be able to edit an item. Properties are handled and validated like in catalogue items, however you cannot edit the `catalogue_item_id`

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes #13

### DIFF
--- a/inventory_management_system_api/repositories/item.py
+++ b/inventory_management_system_api/repositories/item.py
@@ -90,3 +90,18 @@ class ItemRepo:
 
         items = self._items_collection.find(query)
         return [ItemOut(**item) for item in items]
+
+
+    def update(self, item_id: str, item: ItemIn) -> ItemOut:
+        """
+        Update an item by its ID in a MongoDB database.
+
+        :param item_id: The ID of the item to update.
+        :param item: The item containing the update data.
+        :return: The updated item.
+        """
+        item_id = CustomObjectId(item_id)
+        logger.info("Updating item with ID: %s in the database", item_id)
+        self._items_collection.update_one({"_id": item_id}, {"$set": item.model_dump()})
+        item = self.get(str(item_id))
+        return item
--- a/inventory_management_system_api/repositories/item.py
+++ b/inventory_management_system_api/repositories/item.py
@@ -91,7 +91,6 @@ class ItemRepo:
         items = self._items_collection.find(query)
         return [ItemOut(**item) for item in items]
 
-
     def update(self, item_id: str, item: ItemIn) -> ItemOut:
         """
         Update an item by its ID in a MongoDB database.
--- a/inventory_management_system_api/routers/v1/item.py
+++ b/inventory_management_system_api/routers/v1/item.py
@@ -127,7 +127,7 @@ def get_item(
 def partial_update_item(
     item: ItemPatchRequestSchema,
     item_id: Annotated[str, Path(description="The ID of the item to update")],
-    item_service: Annotated[ItemService, Depends()],
+    item_service: Annotated[ItemService, Depends(ItemService)],
 ) -> ItemSchema:
     # pylint: disable=missing-function-docstring
     logger.info("Partially updating item with ID: %s", item_id)
@@ -151,3 +151,7 @@ def partial_update_item(
         message = "An item with such ID was not found"
         logger.exception(message)
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
+    except DatabaseIntegrityError as exc:
+        message = "Unable to update item"
+        logger.exception(message)
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=message) from exc
--- a/inventory_management_system_api/routers/v1/item.py
+++ b/inventory_management_system_api/routers/v1/item.py
@@ -8,11 +8,12 @@ from fastapi import APIRouter, Query, status, HTTPException, Depends, Path
 
 from inventory_management_system_api.core.exceptions import (
     InvalidObjectIdError,
+    MissingMandatoryCatalogueItemProperty,
     MissingRecordError,
     DatabaseIntegrityError,
     InvalidCatalogueItemPropertyTypeError,
 )
-from inventory_management_system_api.schemas.item import ItemPostRequestSchema, ItemSchema
+from inventory_management_system_api.schemas.item import ItemPatchRequestSchema, ItemPostRequestSchema, ItemSchema
 from inventory_management_system_api.services.item import ItemService
 
 logger = logging.getLogger()
@@ -95,4 +96,38 @@ def get_item(
         return ItemSchema(**item.model_dump())
     except InvalidObjectIdError as exc:
         logger.exception("The ID is not a valid ObjectId value")
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
+
+
+@router.patch(path="/{item_id}", summary="Update an item partially by ID", response_description="Item updated successfully",)
+def partial_update_item(
+    item: ItemPatchRequestSchema,
+    item_id: Annotated[str, Path(description="The ID of the item to update")],
+    item_service: Annotated[ItemService, Depends()],
+) -> ItemSchema:
+    # pylint: disable=missing-function-docstring
+    logger.info("Partially updating item with ID: %s", item_id)
+    logger.debug("Item data: %s", item)
+    try:
+        updated_item = item_service.update(item_id, item)
+        return ItemSchema(**updated_item.model_dump())
+    except (InvalidCatalogueItemPropertyTypeError, MissingMandatoryCatalogueItemProperty) as exc:
+        logger.exception(str(exc))
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(exc)) from exc
+    except (MissingRecordError, InvalidObjectIdError) as exc:
+        if (
+            item.catalogue_item_id and item.catalogue_item_id in str(exc) or "catalogue item" in str(exc).lower()
+        ):
+            message = "The specified catalogue item ID does not exist"
+            logger.exception(message)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message)
+        if(
+            item.system_id and item.system_id in str(exc) or "system" in str(exc).lower()
+        ):
+            message = "The specified system ID does not exist"
+            logger.exception(message)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message)
+        
+        message = "An item with such ID was not found"
+        logger.exception(message)
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
--- a/inventory_management_system_api/routers/v1/item.py
+++ b/inventory_management_system_api/routers/v1/item.py
@@ -7,7 +7,7 @@ from typing import Annotated, List, Optional
 from fastapi import APIRouter, Query, status, HTTPException, Depends, Path
 
 from inventory_management_system_api.core.exceptions import (
-    ChildElementsExistError,
+    InvalidActionError,
     InvalidObjectIdError,
     MissingMandatoryCatalogueItemProperty,
     MissingRecordError,
@@ -151,7 +151,7 @@ def partial_update_item(
         message = "Unable to update item"
         logger.exception(message)
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=message) from exc
-    except ChildElementsExistError as exc:
+    except InvalidActionError as exc:
         message = "Cannot change the catalogue item of an item"
         logger.exception(message)
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message) from exc
--- a/inventory_management_system_api/routers/v1/item.py
+++ b/inventory_management_system_api/routers/v1/item.py
@@ -99,7 +99,11 @@ def get_item(
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
 
 
-@router.patch(path="/{item_id}", summary="Update an item partially by ID", response_description="Item updated successfully",)
+@router.patch(
+    path="/{item_id}",
+    summary="Update an item partially by ID",
+    response_description="Item updated successfully",
+)
 def partial_update_item(
     item: ItemPatchRequestSchema,
     item_id: Annotated[str, Path(description="The ID of the item to update")],
@@ -115,19 +119,15 @@ def partial_update_item(
         logger.exception(str(exc))
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(exc)) from exc
     except (MissingRecordError, InvalidObjectIdError) as exc:
-        if (
-            item.catalogue_item_id and item.catalogue_item_id in str(exc) or "catalogue item" in str(exc).lower()
-        ):
+        if item.catalogue_item_id and item.catalogue_item_id in str(exc) or "catalogue item" in str(exc).lower():
             message = "The specified catalogue item ID does not exist"
             logger.exception(message)
             raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message)
-        if(
-            item.system_id and item.system_id in str(exc) or "system" in str(exc).lower()
-        ):
+        if item.system_id and item.system_id in str(exc) or "system" in str(exc).lower():
             message = "The specified system ID does not exist"
             logger.exception(message)
             raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message)
-        
+
         message = "An item with such ID was not found"
         logger.exception(message)
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=message) from exc
--- a/inventory_management_system_api/routers/v1/item.py
+++ b/inventory_management_system_api/routers/v1/item.py
@@ -122,11 +122,11 @@ def partial_update_item(
         if item.catalogue_item_id and item.catalogue_item_id in str(exc) or "catalogue item" in str(exc).lower():
             message = "The specified catalogue item ID does not exist"
             logger.exception(message)
-            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message) from exc
         if item.system_id and item.system_id in str(exc) or "system" in str(exc).lower():
             message = "The specified system ID does not exist"
             logger.exception(message)
-            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message) from exc
 
         message = "An item with such ID was not found"
         logger.exception(message)
--- a/inventory_management_system_api/routers/v1/item.py
+++ b/inventory_management_system_api/routers/v1/item.py
@@ -7,6 +7,7 @@ from typing import Annotated, List, Optional
 from fastapi import APIRouter, Query, status, HTTPException, Depends, Path
 
 from inventory_management_system_api.core.exceptions import (
+    ChildrenElementsExistError,
     InvalidObjectIdError,
     MissingMandatoryCatalogueItemProperty,
     MissingRecordError,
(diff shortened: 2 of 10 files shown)


inventory_management_system_api/repositories/item.py
```diff
@@ -103,3 +103,17 @@ def list(self, system_id: Optional[str], catalogue_item_id: Optional[str]) -> Li
...
+        Update an item by its ID in a MongoDB database.
+
+        :param item_id: The ID of the item to update.
+        :param item: The item containing the update data.
+        :return: The updated item.
+        """
+        item_id = CustomObjectId(item_id)
+        logger.info("Updating item with ID: %s in the database", item_id)
+        self._items_collection.update_one({"_id": item_id}, {"$set": item.model_dump()})
+        item = self.get(str(item_id))
```
joelvdavies, line 118: We also need checks when changing the parent catalogue item/system to at the very least ensure they exist (see catalogue categories/systems). #153 should probably also do the same for catalogue items as it is currently missing.

inventory_management_system_api/services/item.py
```diff
@@ -118,9 +121,69 @@ def get(self, item_id: str) -> Optional[ItemOut]:
...
-
         return self._item_repository.get(item_id)
 
+    def update(self, item_id: str, item: ItemPatchRequestSchema) -> ItemOut:
+        """
+        Update an item by its ID.
+
+        The method checks if the item exists in the database and raises a `MissingRecordError` if it does
+        not. If the catalogue item ID or system ID is being updated, it checks if a catalogue item
+        and/or system ID with such ID exists and raises a `MissingRecordError` if it does not.
```
joelvdavies, line 132: If I understand correctly I think this could do with a message along the lines of
`When updating properties, all existing properties that are different to those of the catalogue item must be provided or they will be overwritten by those in the catalogue item.`

I believe Joshua's edit PR supplies them in one go so this should currently be fine, although I would have initially expected it to just modify those given and leave the rest so that there is no overriding happening here. I would suggest either an explanation like this, or modify it to only replace those given ignoring the contents of the catalogue item. @VKTB any thoughts?


We also need a check like https://github.com/ral-facilities/inventory-management-system/blob/daf48454f38b202392f102aa68e7de8264412061/src/catalogue/items/catalogueItemDirectoryDialog.component.tsx#L83-L86 to prevent moving to a catalogue item with different properties but that is also part of #128 anyway.

joelvdavies (commented):
Haven't looked at the tests yet.

inventory_management_system_api/services/item.py
```diff
@@ -118,9 +121,69 @@ def get(self, item_id: str) -> Optional[ItemOut]:
...
+            if not catalogue_item:
+                catalogue_item = self._catalogue_item_repository.get(stored_item.catalogue_item_id)
+
+            try:
+                catalogue_category_id = catalogue_item.catalogue_category_id
+                catalogue_category = self._catalogue_category_repository.get(catalogue_category_id)
+                if not catalogue_category:
+                    raise DatabaseIntegrityError(f"No catalogue category found with ID: {catalogue_category_id}")
+            except InvalidObjectIdError as exc:
+                raise DatabaseIntegrityError(str(exc)) from exc
```
VKTB, line 169: These are raised here but not caught in the `item` router module.

inventory_management_system_api/services/item.py
```diff
@@ -118,9 +121,69 @@ def get(self, item_id: str) -> Optional[ItemOut]:
...
+                update_data = item.model_dump(exclude_unset=True)
+
+        if "system_id" in update_data and item.system_id != stored_item.system_id:
+            system = self._system_repository.get(item.system_id)
+            if not system:
+                raise MissingRecordError(f"No system found with ID: {item.system_id}")
+
+        if "properties" in update_data:
+            if not catalogue_item:
+                catalogue_item = self._catalogue_item_repository.get(stored_item.catalogue_item_id)
```
VKTB, line 161: Would be nice to have comment(s) here (similar to the ones I left in the `catalogue_item` service module to explain what's going on.

inventory_management_system_api/services/item.py
VKTB: > If I understand correctly I think this could do with a message along the lines of `When updating properties, all existing properties that are different to those of the catalogue item must be provided or they will be overwritten by those in the catalogue item.`

I agree. The `properties` field should also be redeclared in the `ItemPatchRequestSchema` schema model class so that its description can mention this.

> I believe Joshua's edit PR supplies them in one go so this should currently be fine, although I would have initially expected it to just modify those given and leave the rest so that there is no overriding happening here. I would suggest either an explanation like this, or modify it to only replace those given, ignoring the contents of the catalogue item. @VKTB any thoughts?

I don't think the `PATCH` logic of any of the other endpoints update only those provided(?) so I would rather not do something different here as it would be confusing.

VKTB (changes requested):
Will review the tests after the requested changes are addressed.

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    updated_item = item_service.update(
+        item.id, ItemPatchRequestSchema(is_defective=item.is_defective, usage_status=item.usage_status)
+    )
+
+    item_repository_mock.update.assert_called_once_with(
+        item.id,
+        ItemIn(
+            catalogue_item_id=item.catalogue_item_id,
+            system_id=item.system_id,
+            **FULL_ITEM_INFO,
```
joelvdavies, line 339:
```suggestion
            **FULL_ITEM_INFO,
            catalogue_item_id=item.catalogue_item_id,
            system_id=item.system_id,
```

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    assert str(exc.value) == f"No item found with ID: {item_id}"
+
+
+def test_update_change_catalogue_item_id(
+    test_helpers, item_repository_mock, catalogue_item_repository_mock, catalogue_category_repository_mock, item_service
+):
+    """
+    Test updating catalogue item id to an existing id
+    """
+    item = ItemOut(id=str(ObjectId()), catalogue_item_id=str(ObjectId()), system_id=str(ObjectId()), **FULL_ITEM_INFO)
```
joelvdavies, line 366:
```suggestion
    item = ItemOut(**FULL_ITEM_INFO, id=str(ObjectId()), catalogue_item_id=str(ObjectId()), system_id=str(ObjectId()), **FULL_ITEM_INFO)
```

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
     item_repository_mock.get.assert_called_once_with(item_id)
+
+
+def test_update(test_helpers, item_repository_mock, item_service):
+    """
+    Test updating an item,
+
+    Verify that the `update` method properly handles the item to be updated.
+    """
+    item = ItemOut(id=str(ObjectId()), catalogue_item_id=str(ObjectId()), system_id=str(ObjectId()), **FULL_ITEM_INFO)
```
joelvdavies, line 320:
```suggestion
    item = ItemOut(**FULL_ITEM_INFO, id=str(ObjectId()), catalogue_item_id=str(ObjectId()), system_id=str(ObjectId()))
```

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    manufacturer_id = str(ObjectId())
+
+    # Mock `get` to return a catalogue item
+    test_helpers.mock_get(
+        catalogue_item_repository_mock,
+        CatalogueItemOut(
+            id=item.catalogue_item_id,
+            catalogue_category_id=catalogue_category_id,
+            manufacturer_id=manufacturer_id,
+            **FULL_CATALOGUE_ITEM_A_INFO,
```
joelvdavies, line 384:
```suggestion
            **FULL_CATALOGUE_ITEM_A_INFO,
            id=item.catalogue_item_id,
            catalogue_category_id=catalogue_category_id,
            manufacturer_id=manufacturer_id,
```

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+        ),
+    )
+    # Mock `get` to return a catalogue category
+    test_helpers.mock_get(
+        catalogue_category_repository_mock,
+        CatalogueCategoryOut(id=catalogue_category_id, **FULL_CATALOGUE_CATEGORY_A_INFO),
```
joelvdavies, line 390:
```suggestion
        CatalogueCategoryOut(**FULL_CATALOGUE_CATEGORY_A_INFO, id=catalogue_category_id),
```

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    # Mock `update` to return the updated item
+    test_helpers.mock_update(item_repository_mock, item)
+
+    updated_item = item_service.update(
+        item.id,
+        ItemPatchRequestSchema(catalogue_item_id=item.catalogue_item_id),
+    )
+
+    item_repository_mock.update.assert_called_once_with(
+        item.id, ItemIn(catalogue_item_id=item.catalogue_item_id, system_id=item.system_id, **FULL_ITEM_INFO)
```
joelvdavies, line 402:
```suggestion
        item.id, ItemIn(**FULL_ITEM_INFO, catalogue_item_id=item.catalogue_item_id, system_id=item.system_id)
```

test/unit/services/test_item.py
```diff
@@ -304,3 +309,344 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    assert updated_item == item
+
+
+def test_update_with_nonexistent_catalogue_item_id(
+    test_helpers, catalogue_item_repository_mock, item_repository_mock, item_service
+):
+    """
+    Test updating an item with a non-existent catalogue item ID.
+    """
+    item = ItemOut(id=str(ObjectId()), catalogue_item_id=str(ObjectId()), system_id=str(ObjectId()), **FULL_ITEM_INFO)
```
joelvdavies, line 413:
```suggestion
    item = ItemOut(**FULL_ITEM_INFO, id=str(ObjectId()), catalogue_item_id=str(ObjectId()), system_id=str(ObjectId()))
```

joelvdavies: There are more of these below as well.

joelvdavies (commented):
Could the **FULL_ITEM_INFO's go at the start when of these to ensure they are actually overwritten? I have only commented on some of them.

inventory_management_system_api/services/item.py
```diff
@@ -118,9 +121,75 @@ def get(self, item_id: str) -> Optional[ItemOut]:
...
+
+        catalogue_item = None
+        if "catalogue_item_id" in update_data and item.catalogue_item_id != stored_item.catalogue_item_id:
+            catalogue_item = self._catalogue_item_repository.get(item.catalogue_item_id)
+            if not catalogue_item:
+                raise MissingRecordError(f"No catalogue item found with ID: {item.catalogue_item_id}")
+
+            if "properties" not in update_data:
+                item.properties = [PropertyPostRequestSchema(**prop.model_dump()) for prop in stored_item.properties]
+                update_data = item.model_dump(exclude_unset=True)
```
VKTB, line 154: It was confirmed yesterday that this case won't occur so we should not allow for an item to be moved from one catalogue item to another. I wonder if it is possible in pydantic to remove the inherited `catalogue_item_id` property from the `ItemPatchRequestSchema` model. If that's possible then that can be one solution. Another solution would be to raise an exception here. @joelvdavies thoughts?

Once you address this, please remove or update any tests that test this logic.

inventory_management_system_api/routers/v1/item.py
```diff
@@ -116,3 +117,41 @@ def get_item(
...
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(exc)) from exc
+    except (MissingRecordError, InvalidObjectIdError) as exc:
+        if item.catalogue_item_id and item.catalogue_item_id in str(exc) or "catalogue item" in str(exc).lower():
+            message = "The specified catalogue item ID does not exist"
+            logger.exception(message)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message) from exc
+        if item.system_id and item.system_id in str(exc) or "system" in str(exc).lower():
+            message = "The specified system ID does not exist"
+            logger.exception(message)
+            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=message) from exc
```
VKTB, line 149: e2e tests to cover these cases would be nice.

test/e2e/test_item.py
VKTB, line 1: Would be nice to also test when `properties` are updated. Testing when not all existing properties are supplied would too be nice.

VKTB (changes requested):

inventory_management_system_api/services/item.py
joelvdavies: Ah sorry only just saw this comment, I prefer the current solution though - raising an error as I think removing inherited variables could be a bit confusing and I don't think it is possible without using another base class.

test/e2e/test_item.py
```diff
@@ -601,3 +602,254 @@ def test_get_items_with_invalid_catalogue_item_id_filter(test_client):
...
+
+    item_patch = {"usage_status": 1, "is_defective": True}
+
+    response = test_client.patch(f"/v1/items/{item_id}", json=item_patch)
+
+    assert response.status_code == 200
+
+    item = response.json()
+
+    assert item == {**ITEM_POST_EXPECTED, **item_patch, "catalogue_item_id": catalogue_item_id, "system_id": system_id}
```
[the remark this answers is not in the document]
VKTB: @MatteoGuarnaccia5 Unresolving this as I don't think this has been addressed for the other tests. `test_partial_update_change_system_id`, `test_partial_update_change_nonexistent_system_id` etc are some to name.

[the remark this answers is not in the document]
VKTB: I addressed it as part of https://github.com/ral-facilities/inventory-management-system-api/pull/152/commits/0e4a2ccaaec2918ffccba01c6aa1cf9f0aaa2559 for you.

test/e2e/test_item.py
```diff
@@ -601,3 +602,360 @@ def test_get_items_with_invalid_catalogue_item_id_filter(test_client):
...
+    item_post = {**ITEM_POST, "catalogue_item_id": catalogue_item_id, "system_id": system_id}
+    response = test_client.post("/v1/items", json=item_post)
+
+    item_patch = {
+        "properties": [
+            {"name": "Property A", "value": 12, "unit": "mm"},
+            {"name": "Property B", "value": True, "unit": None},
+            {"name": "Property C", "value": "20x15x10", "unit": "cm"},
+        ],
+    }
```
VKTB, line 805: Units should not be supplied in `PATCH` requests. If you look at the definition of `PropertyPostRequestSchema`, it only accepts `name` and `value`.

test/e2e/test_item.py
```diff
@@ -750,6 +742,120 @@ def test_partial_update_change_system_id(test_client):
...
+    response = test_client.post("/v1/catalogue-items", json=catalogue_item_post)
+    catalogue_item_id = response.json()["id"]
+
+    # pylint: enable=duplicate-code
+    item_post = {**ITEM_POST, "catalogue_item_id": catalogue_item_id, "system_id": system_id}
+    response = test_client.post("/v1/items", json=item_post)
+
+    item_patch = {
+        "properties": [{"name": "Property A", "value": 15, "unit": "mm"}],
+    }
```
VKTB, line 841: I think it would be better if this supplies `Property B` and `Property C` instead of `Property A`. The value of `Property A` at a catalogue item level is `20` and the test creates an item and overrides the value of this property at an item level to `20`. So if we don't supply `Property A` in the `PATCH` request, it should technically inherit the value from the catalogue item and become `21, which is exactly what we need to verify.

`Property A` and `Property B` are not overridden when the item is created so it is expected that they will stay the same after the `PATCH`.

test/e2e/test_item.py
```diff
@@ -601,3 +602,360 @@ def test_get_items_with_invalid_catalogue_item_id_filter(test_client):
...
+    response = test_client.patch(f"/v1/items/{response.json()['id']}", json=item_patch)
+
+    assert response.status_code == 200
+
+    item = response.json()
+
+    assert item == {**ITEM_POST_EXPECTED, **item_patch, "catalogue_item_id": catalogue_item_id, "system_id": system_id}
+
+
+def test_partial_update_not_all_supplied_properties(test_client):
```
VKTB, line 815:
```suggestion
def test_partial_update_with_missing_existing_properties(test_client):
```

test/e2e/test_item.py
```diff
@@ -601,3 +602,360 @@ def test_get_items_with_invalid_catalogue_item_id_filter(test_client):
...
+    """
+    item_patch = {"usage_status": 1, "is_defective": True}
+
+    response = test_client.patch(f"/v1/items/{str(ObjectId())}", json=item_patch)
+
+    assert response.status_code == 404
+    assert response.json()["detail"] == "An item with such ID was not found"
+
+
+def test_partial_update_try_change_catalogue_item_id(test_client):
```
VKTB, line 667:
```suggestion
def test_partial_update_change_catalogue_item_id(test_client):
```

test/unit/services/test_item.py
```diff
@@ -304,3 +317,369 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    # Mock `get` to return an item
+    test_helpers.mock_get(item_repository_mock, None)
+
+    item_id = str(ObjectId())
+    with pytest.raises(MissingRecordError) as exc:
+        item_service.update(item_id, ItemPatchRequestSchema(properties=[]))
+    assert str(exc.value) == f"No item found with ID: {item_id}"
+
+
+def test_try_to_update_catalogue_item_id(test_helpers, item_repository_mock, item_service):
```
VKTB, line 375:
```suggestion
def test_update_change_catalogue_item_id(test_helpers, item_repository_mock, item_service):
```

test/unit/services/test_item.py
```diff
@@ -304,3 +317,369 @@ def test_get_with_nonexistent_id(test_helpers, item_repository_mock, item_servic
...
+    properties[0]["value"] = "20"
+    with pytest.raises(InvalidCatalogueItemPropertyTypeError) as exc:
+        item_service.update(
+            item.id,
+            ItemPatchRequestSchema(properties=properties),
+        )
+    assert str(exc.value) == "Invalid value type for catalogue item property 'Property A'. Expected type: number."
+
+
+def test_update_change_valuie_for_boolean_property_invalid_type(
```
VKTB, line 640:
```suggestion
def test_update_change_value_for_boolean_property_invalid_type(
```

test/unit/services/test_item.py
VKTB, line 1: A unit test to cover the services logic for missing existing properties would be nice here too.

inventory_management_system_api/services/item.py
```diff
@@ -118,9 +122,67 @@ def get(self, item_id: str) -> Optional[ItemOut]:
...
+        :return: The updated item.
+        """
+        update_data = item.model_dump(exclude_unset=True)
+
+        stored_item = self.get(item_id)
+        if not stored_item:
+            raise MissingRecordError(f"No item found with ID: {item_id}")
+
+        if "catalogue_item_id" in update_data and item.catalogue_item_id != stored_item.catalogue_item_id:
+            raise ChildElementsExistError("Cannot change the catalogue item the item belongs to")
```
VKTB, line 148: This exception is not relevant for this case. I discussed this with Joel and we think that it would be better if we create a new custom exception `InvalidActionError`, raise that here and handle it in the endpoint.

VKTB (changes requested):
Sorry a few more minor things and this should be ready to go in.

inventory_management_system_api/services/item.py
MatteoGuarnaccia5: what http status code should I use for it? still 422?

VKTB: I think so. We can change it later if we are proven wrong as it is usually the case with HTTP status codes 😅

test/e2e/test_item.py
```diff
@@ -601,3 +609,372 @@ def test_get_items_with_invalid_catalogue_item_id_filter(test_client):
...
+    assert response.status_code == 200
+
+    item = response.json()
+
+    assert item == {
+        **ITEM_POST_EXPECTED,
+        "properties": [
+            {"name": "Property B", "unit": None, "value": False},
+            {"name": "Property C", "unit": "cm", "value": "25x10x5"},
+            {"name": "Property A", "value": 20, "unit": "mm"},
```
VKTB, line 868: The properties get reordered when sending a `PATCH` request to the item endpoint with missing existing properties. Is this fine for the frontend @MatteoGuarnaccia5 @joelvdavies @joshuadkitenge?

joelvdavies: Just checked edit, yes it does effect it if you delete properties in the create or edit dialogues, ideally they would be kept in the same order. I wouldn't mind it being addressed in a separate PR though if that is the only issue you found.

joelvdavies: @joshuadkitenge should we perhaps have placeholder text on the dialog textboxes so that when they are empty they display the default value to make it clearer what will happen?

VKTB (approved):

joelvdavies (approved):

PR #163, opened 2024-01-26
Preserve the order of properties when creating or updating items #162

### LINKED ISSUE #162: Preserve order of properties in items
Currently when creating/editing an item and missing out say the first property, the default from the catalogue item gets appended onto the end, rather than preserve the order defined in the catalogue item. For the front end it would be best if it could be kept in order so all items created/updated appear the same.

See https://github.com/ral-facilities/inventory-management-system-api/pull/152#discussion_r1466647025. I think we may also be missing a test in the create that would have picked this up.

joelvdavies:
## Description
Preserves the order of properties defined in the catalogue item when creating/updating an item.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #162

### DIFF
--- a/inventory_management_system_api/services/item.py
+++ b/inventory_management_system_api/services/item.py
@@ -75,15 +75,8 @@ class ItemService:
             raise DatabaseIntegrityError(str(exc)) from exc
 
         supplied_properties = item.properties if item.properties else []
-        missing_supplied_properties = self._find_missing_supplied_properties(
-            catalogue_item.properties, supplied_properties
-        )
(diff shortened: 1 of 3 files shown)


MatteoGuarnaccia5 (approved):
LGTM

VKTB (approved):

VKTB:
I am assuming that we do not have this ordering issue with catalogue items?

joelvdavies:
> I am assuming that we do not have this ordering issue with catalogue items?

No we don't the properties are just overridden at the moment, so it doesn't have the same issue when editing, and when creating the optional fields just aren't stored.

PR #169, opened 2024-01-30
Secure API Endpoints with JWT Access Token

### LINKED ISSUE #164: Secure API Endpoints with JWT Access Token
At present, the API endpoints are accessible without any authentication, meaning that any user can consume them. Implementing JWT access token protection for our API endpoints will ensure that only authenticated and authorized users can access the resources. This approach adds a layer of security by validating the authenticity of incoming requests, mitigating the risk of unauthorized access.

The proposed solution is to integrate the [LDAP-JWT Authentication Service](https://github.com/ral-facilities/ldap-jwt-auth) into the API so that it only allows access to the resources to JWT access tokens generated by this service. The API should extract the JWT access token supplied in the `Authorization` header of the requests, validate and decode the JWT access token, and decide whether the user making the request is authorised to access the resource. For now, decoding and checking that the JWT access token has not expired is enough but in future, we will need to implement user roles so that we can perform specific checks on the claims embedded in the token to determine whether the requesting user has the necessary permissions.

VKTB:
## Description
Implements JWT access token protection for the API endpoints. This ensures that only authenticated and authorised users can access the resources. It achieves this by integrating the [LDAP-JWT Authentication Service](https://github.com/ral-facilities/ldap-jwt-auth) into the API, allowing access to the resources to JWT access tokens generated by this service only. The API extracts the JWT access token supplied in the `Authorization` header of the requests, validates and decodes the JWT access token, and decides whether the user making the request is authorised to access the resource. For now, decoding and checking that the JWT access token has not expired is enough but in future, we will need to implement user roles so that we can perform specific checks on the claims embedded in the token to determine whether the requesting user has the necessary permissions. The JWT access tokens are decoded using the public key used in the LDAP-JWT Authentication Service.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #164


VKTB:
This would require changes in the frontend before it can be merged.

joelvdavies:
> This would require changes in the frontend before it can be merged.

Could this have a environment variable for development or something to disable it - running IMS alone in dev without SciGateway presumably wouldn't work. I don't think it would need to wait for front end changes with this - the only thing is the SciGateway integration/error messages but those aren't breaking.

VKTB:
Had to downgrade `pytest` to `7.4.4` as `pytest-asyncio` `0.23.4` depends on `pytest``<8` and `>=7.0.0`.

inventory_management_system_api/core/config.py
```diff
@@ -17,6 +19,34 @@ class APIConfig(BaseModel):
...
+    Configuration model for the JWT access token authentication/authorization.
+    """
+
+    enabled: bool = False
+    public_key_path: Optional[str] = Field(default=None, validate_default=True)
+    jwt_algorithm: Optional[str] = Field(default=None, validate_default=True)
+
+    @field_validator("public_key_path", "jwt_algorithm")
+    @classmethod
+    def validate_unit(cls, field_value: str, info: ValidationInfo) -> Optional[str]:
```
joelvdavies, line 33: This function name should be changed.

inventory_management_system_api/core/config.py
```diff
@@ -17,6 +19,34 @@ class APIConfig(BaseModel):
...
+        """
+        Validator for the `public_key_path` and `jwt_algorithm` fields to make them mandatory if the value of the
+        `enabled` is `True`
+
+        It checks if the `enabled` field has been set to `True` and raises a `TypeError` if this is the case.
+
+        :param field_value: The value of the field.
+        :param info: Validation info from pydantic.
+        :return: The value of the field.
+        :raises ValueError: If `unit` is provided when `type` is set to `boolean`.
```
joelvdavies, line 43: Also this comment.

inventory_management_system_api/core/consts.py
```diff
@@ -1,8 +1,19 @@
...
 # Maximum length of the trail to return for breadcrumbs GET endpoints
 # e.g. a value of 5 means the trail goes back to at most the last 4
 # parents
 BREADCRUMBS_TRAIL_MAX_LENGTH: int = 5
+
+if config.authentication.enabled:
+    # Read the content of the public key file into a constant. This is used for decoding of JWT access tokens.
+    try:
+        with open(config.authentication.public_key_path, "r", encoding="utf-8") as f:
+            PUBLIC_KEY = f.read()
```
joelvdavies, line 17:
```suggestion
        with open(config.authentication.public_key_path, "r", encoding="utf-8") as file:
            PUBLIC_KEY = file.read()
```

test/e2e/test_jwt_bearer.py
```diff
@@ -0,0 +1,74 @@
...
+            assert response.status_code == 403
+            assert response.json()["detail"] == expected_response_message
+        if "PATCH" in api_route.methods:
+            response = test_client.patch(api_route.path, headers=headers)
+            assert response.status_code == 403
+            assert response.json()["detail"] == expected_response_message
+        if "POST" in api_route.methods:
+            response = test_client.post(api_route.path, headers=headers)
+            assert response.status_code == 403
+            assert response.json()["detail"] == expected_response_message
```
joelvdavies, line 74: Could this not just be a for loop. Its probably worth putting PUT here as well.

joelvdavies (changes requested):
A few minor things and a request to get vscode's tests to run properly.

.github/workflows/.ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         run: cp inventory_management_system_api/logging.example.ini inventory_management_system_api/logging.ini
 
       - name: Run unit tests
-        run: pytest test/unit/ --cov
+        run: AUTHENTICATION__PUBLIC_KEY_PATH="./test/keys/jwt-key.pub" pytest test/unit/ --cov
```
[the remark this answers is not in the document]
VKTB: Didn't realise that env variables can be specified in `pytest.ini`, it makes life so much easier, thanks!

inventory_management_system_api/core/config.py
VKTB: Oops, I forgot to change this after pasting it.

inventory_management_system_api/core/config.py
VKTB: Oops, I forgot to change this after pasting it.

test/e2e/test_jwt_bearer.py
VKTB: It can be yeah, thanks.

VKTB:
Please note that after overriding the env var values used during testing by using pytest.ini, the tests should now be run with the following commands:
- Unit tests -> `pytest -c test/pytest.ini test/unit/`
- E2E tests -> `pytest -c test/pytest.ini test/e2e/`
- All tests -> `pytest -c test/pytest.ini test`

README was updated accordingly.

README.md
```diff
@@ -19,15 +19,24 @@ not be used in production.
...
+    ```
+
+2. Create a `logging.ini` file alongside the `logging.example.ini` file. Use the example file as a reference and modify it accordingly:
     ```bash
-   cp logging.example.ini logging.ini
+    cp logging.example.ini logging.ini
     ```
 
-2. Build and start the Docker containers:
+3. (**Required only if JWT Auth is enabled**) Create a `keys` directory in the root of the project directory and inside it create a copy of the public key generated by the authentication component. This is needed for decoding of JWT access tokens signed by the corresponding private key.
```
VKTB, line 32: `keys` dir has now been committed so there is no need to be created.
```suggestion
3. (**Required only if JWT Auth is enabled**) Inside the `keys` directory at the root of the project directory, create a copy of the public key generated by the authentication component. This is needed for decoding of JWT access tokens signed by the corresponding private key.
```

keys/.keep
MatteoGuarnaccia5, line 1: Following our discussion this folder needs to be removed right? I think you told me to make a note of it

README.md
```diff
@@ -19,15 +19,24 @@ not be used in production.
...
+    ```
+
+2. Create a `logging.ini` file alongside the `logging.example.ini` file. Use the example file as a reference and modify it accordingly:
     ```bash
-   cp logging.example.ini logging.ini
+    cp logging.example.ini logging.ini
     ```
 
-2. Build and start the Docker containers:
+3. (**Required only if JWT Auth is enabled**) Inside the `keys` directory at the root of the project directory, create a copy of the public key generated by the authentication component. This is needed for decoding of JWT access tokens signed by the corresponding private key.
```
MatteoGuarnaccia5, line 32: Could probably mention as well to ensure there is a `.keep` file, otherwise don't people risk just losing their public key as it is in the `.gitignore` file

joelvdavies: I don't think this is an issue, the .keep just ensures the folder is committed if git thinks its empty, which it will do since the jwt is in the git ignore. We don't want anyone committing their jwt public key accidentally.

keys/.keep
joelvdavies: @VKTB Why do we commit the keys folder?

joelvdavies (commented):
Worked well for me, tested against https://github.com/ral-facilities/inventory-management-system/pull/275, see comment https://github.com/ral-facilities/inventory-management-system/pull/275#pullrequestreview-1858832558.

README.md
joelvdavies, line 32: This should mention creating the keys directory like ldap-jwt-auth as it is no longer committed.

VKTB: Sorry I forgot to revert the changes for this. Addressed now in https://github.com/ral-facilities/inventory-management-system-api/pull/169/commits/98866f6433220d558752f7356775ae96515d920c.

joelvdavies (approved):

PR #172, opened 2024-02-01
Allow assignment of None to optional catalogue item & item properties #171

### LINKED ISSUE #171: Giving optional item property value of None says invalid type
Caused by https://github.com/ral-facilities/inventory-management-system-api/blob/67446ce2d9d0ff7b69cd99aa54d8ff28f748bd52/inventory_management_system_api/services/utils.py#L123

Previously was fine as we only sent back values if they were given by the user.

joelvdavies:
## Description

Allows items and catalogue items to have non-mandatory properties assigned a value of None rather than just relying on them not being defined. This resolves an issue on the front end where properties can't be set to None as passing undefined just obtains them from the catalogue item instead.

Also discovered #173  in testing, so added some TODO's.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #171


inventory_management_system_api/services/utils.py
```diff
@@ -116,24 +116,33 @@ def _validate_catalogue_item_property_value(defined_property: Dict, supplied_pro
 
     defined_type = defined_property["type"]
     defined_allowed_values = defined_property["allowed_values"]
+    defined_value_mandatory = defined_property["mandatory"]
```
VKTB, line 119: I am not sure about the usage of `value` in this one, maybe `defined_property_mandatory`? The two above may need to change as well to keep things consistent `defined_property_type` etc.

VKTB (approved):

PR #174, opened 2024-02-02
Fix item properties being overridden when the property is non-mandatory  #173

### LINKED ISSUE #173: Item properties not given a value if its non-mandatory in catalogue item
When a catalogue item has a non mandatory property that isnt given when it is created, that value is completely ignored at the catalogue item level regardless of what it is assigned.

To fix this  `_filter_matching_catalogue_item_properties` in utils should be modified to be like `_merge_missing_properties`, with a name e.g. `_merge_non_mandatory_properties`, merging any unsupplied non-mandatory properties with those from the catalogue category with a default value of `None`.

joelvdavies:
## Description
See #173. This PR depends on #172.

These changes mean that all properties are stored for a catalogue item even when they were specified to be none, and are also returned in responses. This in turn fixes those properties being ignored when a value is supplied for them during an item create/update request.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #173


inventory_management_system_api/services/utils.py
```diff
@@ -184,22 +184,20 @@ def _check_missing_mandatory_catalogue_item_properties(
...
-    matching_properties = {}
-    for supplied_property_name, supplied_property in supplied_properties.items():
-        if supplied_property_name in defined_properties:
-            matching_properties[supplied_property_name] = supplied_property
+    logger.info("Merging any missing defined non-mandatory properties with the supplied properties")
 
-    return matching_properties
+    properties: Dict[str, Dict] = {}
+    for defined_property_name, defined_property in defined_properties.items():
+        supplied_property = supplied_properties.get(defined_property_name)
```
VKTB, line 198: Does `get` return `None` if it cannot find the provided key even if `None` isn't provided as the default return value?

inventory_management_system_api/services/utils.py
```diff
@@ -184,22 +184,20 @@ def _check_missing_mandatory_catalogue_item_properties(
...
 ) -> Dict[str, Dict]:
     """
-    Filter through the supplied properties and extract the ones matching the defined properties.
-
-    :param defined_properties: The defined catalogue item properties stored as part of the catalogue category in the
-        database.
-    :param supplied_properties: The supplied catalogue item properties.
-    :return: The supplied properties that are matching the defined properties.
+    Merges in any non-mandatory properties that have not been supplied, giving them a value of None, using
+    the same order as they are defined.
```
VKTB, line 192: Can we have a note here to say that any extra undefined properties that get supplied are ignored?

VKTB (commented):
Will these changes require changes in the frontend given the catalogue item responses now include all properties?

VKTB (approved):

PR #181, opened 2024-02-07
Add units endpoint #170

### LINKED ISSUE #170: Add units endpoint
See operationsgateway-api users for how we can populate the initial units.

joelvdavies:
## Description
Adds a `/units` endpoint for obtaining a list of available units.

Used by https://github.com/ral-facilities/inventory-management-system/pull/307.

See readme for instructions to import the units.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #170


data/units.json
VKTB, line 1: The `_id`s get imported as strings in the database rather than as `ObjectId`s. I think `$oid` needs to be specified for the `_id`s.

inventory_management_system_api/routers/v1/unit.py
```diff
@@ -0,0 +1,25 @@
...
+
+from inventory_management_system_api.schemas.unit import UnitSchema
+from inventory_management_system_api.services.unit import UnitService
+
+logger = logging.getLogger()
+
+router = APIRouter(prefix="/v1/units", tags=["units"])
+
+
+@router.get(path="/", summary="Get Units", response_description="List of Units")
```
VKTB, line 19: I take it that we don't need an endpoint for getting a unit by ID?

VKTB (changes requested):
Just one minor thing to do with the way ID get imported/stored but works well otherwise.

inventory_management_system_api/routers/v1/unit.py
joelvdavies: Yes we don't. I will remove the  "/" here as well.

VKTB (approved):

PR #184, opened 2024-02-09
Prevent catalogue items moving to category when the properties are not identical #128

### LINKED ISSUE #128: Allow catalogue items to be moved between categories only if the catalogue item properties are fully matching

joelvdavies:
## Description
Only allows moving a catalogue item to a category with properties that exactly match (including units and order), when the properties are not supplied. If the properties are supplied, the usual checks apply i.e. it can be moved provided the new properties are valid for the place it is moving to, and that the catalogue item doesn't have any children.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #128


test/e2e/test_catalogue_item.py
```diff
@@ -1048,10 +1048,46 @@ def test_partial_update_catalogue_item_change_catalogue_category_id_without_prop
     assert response.status_code == 422
     assert (
         response.json()["detail"]
-        == "Invalid value type for catalogue item property 'Property A'. Expected type: boolean."
+        == "Cannot move catalogue item to a category with different catalogue_item_properties without specifying the "
+        "new properties"
     )
 
 
+def test_partial_update_catalogue_item_change_catalogue_category_id_with_properties(test_client):
```
VKTB, line 1056: Isn't this a duplicate of `test_partial_update_catalogue_item_change_catalogue_category_id`?

test/e2e/test_catalogue_item.py
VKTB, line 1: Maybe a test where the order is different?

VKTB (approved):

Issue #90, opened 2023-10-18
Clean up tests


VKTB:
- [ ] Refactor tests where possible
- [ ] Remove code duplication

etc

joelvdavies:
I think it would be nice to combine post requests in unit tests e.g. a utility function something like
```
def _post_catalogue_categories(test_client, catalogue_categories: list[dict], manufacturer: Optional[dict] = None):
    """Utility function that posts a catalogue categories and a manufacturer and returns a list of
    corresponding dictionaries containing the ids ready for passing to a catalogue item

    :param catalogue_categories: Catalogue category dictionaries to post (When None uses the default)
    :param manufacturer: Manufactuer dictionary to post (When None uses the default)
    :returns: List of dictionary containing the catalogue category id and manufacturer id for each given
              category dictionary
    """
    id_dicts = []

    manufacturer_id = None
    if manufacturer is not None:
        response = test_client.post("/v1/manufacturers", json=manufacturer)
        manufacturer_id = response.json()["id"]

    for catalogue_category in catalogue_categories:
        response = test_client.post("/v1/catalogue-categories", json=catalogue_category)
        id_dict = {"catalogue_category_id": response.json()["id"]}
        if manufacturer_id is not None:
            id_dict["manufacturer_id"] = manufacturer_id
        id_dicts.append(id_dict)

    return id_dicts
```
Could be used in catalogue_items so avoid posting the categories and manufacturer's repeatedly, items likewise repeats posting manufacturers, catalogue categories and catalogue items but with functions like these they could be imported and added upon.

VKTB:
A util class like the one below might be nice

```python
class TestClient:

    def __init__(self, test_client):
        # self.catalogue_category = {
        #     "id": None,
        #     "parent_id": None,
        #     "name": None,
        #     "code": None,
        #     "is_leaf": None,
        #     "catalogue_item_properties": None,
        # }
        self.test_client = test_client
        self.catalogue_categories = {}
        self.catalogue_items = {}
        self.catalogue_items_expected = {}
        self.items = {}
        self.manufacturers = {}
        self.systems = {}

    def create_catalogue_category(self, catalogue_category_data: dict):
        response = self.test_client.post("/v1/catalogue-categories", json=catalogue_category_data)
        catalogue_category = response.json()
        self.catalogue_categories[catalogue_category["id"]] = catalogue_category
        return response.status_code, catalogue_category

[...]

PR #193, opened 2024-02-16
Add creation modified times #100

### LINKED ISSUE #100: Add creation and modified times
Likely for every entity stored in the database e.g. catalogue categories, catalogue items, items, systems

joelvdavies:
## Description
Adds `created_time` and `updated_time` fields to catalogue categories, catalogue items, items, manufacturers and systems.

Draft as tests need updating.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #100


test/unit/services/test_catalogue_item.py
```diff
@@ -1202,7 +1334,11 @@ def test_update_with_obsolete_replacement_catalogue_item_id(
             id=str(ObjectId()),
```
VKTB, line 1334: Should ID be different here? Can `model_dump()` be used?

test/unit/services/test_catalogue_item.py
```diff
@@ -1327,19 +1473,30 @@ def test_update_add_non_mandatory_property(
...
+            **{
+                **FULL_CATALOGUE_ITEM_A_INFO,
+                "created_time": catalogue_item.created_time,
+            },
         ),
     )
     assert updated_catalogue_item == catalogue_item
 
 
 def test_update_remove_non_mandatory_property(
```
VKTB, line 1485: Is this test working properly? `catalogue_item` has 2 props but 3 are provided in `assert_called_once_with`

test/unit/services/test_item.py
```diff
@@ -475,31 +524,46 @@ def test_update_with_nonexistent_system_id(test_helpers, system_repository_mock,
...
-        ItemOut(**{**FULL_ITEM_INFO, **item.model_dump(), "properties": FULL_ITEM_INFO["properties"]}),
+        ItemOut(
+            **{
+                **FULL_ITEM_INFO,
+                **item.model_dump(),
+                "properties": FULL_ITEM_INFO["properties"],
+                "created_time": item.created_time,
+                "modified_time": item.created_time,
+            }
+        ),
```
VKTB, line 566: Is the merging order correct here?

Same question for line 633.

VKTB (commented):
I noticed the following before while fixing the service unit tests.

test/unit/services/test_item.py
joelvdavies: Yes I think that makes sense - its the most general first, and any extra at the end.

test/unit/services/test_catalogue_item.py
joelvdavies: No it should be the same, I think model dump can be used, but it would also need `is_obsolete` to be changed and ideally the `obsolete_replacement_catalogue_item_id` would also be None, so I don't think it would be worth it here as you would be dumping the expected output of the test rather than an initial model.

test/unit/services/test_catalogue_item.py
VKTB: Sorry by props I meant `properties`. `catalogue_item`, which is used to assert against the `updated_catalogue_item`, only has two `properties` but three `properties` are provided to the `assert_called_once_with` for the `update` repo method. If three `properties` are sent to the database then I am assuming three should be returned so `catalogue_item` should have three?

joelvdavies: Yes that does make sense to me due to https://github.com/ral-facilities/inventory-management-system-api/pull/174 - its adding in the non mandatory one back in but with a value of `None`. The order on line 1512/1513 should be reversed though considering its a dictionary.

test/unit/services/test_item.py
joelvdavies: Yeah you can do its just a case of keeping the same sort of logic across tests, but its really a case for #90 anyway, all can be broken down to the same overrides but yes for some tests not all will be needed such as this one.

joelvdavies (commented):
Looked through the updated service unit tests, just had a single thing that looked inconsistent, the rest LGTM.

test/unit/services/test_system.py
```diff
@@ -184,25 +209,49 @@ def test_list(system_repository_mock, system_service):
...
     test_helpers.mock_get(
         system_repository_mock,
-        SystemOut(id=system.id, **{**SYSTEM_A_INFO_FULL, "name": "Different name", "code": "different-name"}),
+        SystemOut(
+            id=system.id,
+            **{
+                **SYSTEM_A_INFO_FULL,
+                "name": "Different name",
+                "code": "different-name",
+                "created_time": SYSTEM_A_INFO_FULL["created_time"] - timedelta(days=5),
```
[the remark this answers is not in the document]
VKTB: Yeah, it should be, just an oversight on my part. Thanks!

VKTB (approved):
I reviewed your changes and they look good to me, thanks!

joelvdavies:
> I reviewed your changes and they look good to me, thanks!

I have also reviewed yours and they look good to me too (I can't approve as its my PR)

VKTB (approved):

test/unit/services/test_catalogue_item.py
```diff
@@ -604,15 +635,23 @@ def test_update_when_no_child_elements(test_helpers, catalogue_item_repository_m
         id=str(ObjectId()),
         catalogue_category_id=str(ObjectId()),
         manufacturer_id=str(ObjectId()),
-        **FULL_CATALOGUE_ITEM_A_INFO,
+        **{
+            **FULL_CATALOGUE_ITEM_A_INFO,
+            "created_time": FULL_CATALOGUE_ITEM_A_INFO["created_time"] - timedelta(days=5),
```
MatteoGuarnaccia5, line 640: why is there this `- timedelta(days=5)` in a lot of the tests? I don't understand what its purpose is

MatteoGuarnaccia5 (approved):
Works well for me and all the tests look fine to me as well

test/unit/services/test_catalogue_item.py
VKTB: Both create and update times are generated when something is created. In the update tests, we return what the entity looks like before it is updated so we set the create and modified time 5 days earlier so that we can then ensure that the modified time gets updated to the mocked datetime. Let me know if this does not make sense and I will talk to you.

PR #231, opened 2024-04-11
Add usage statuses endpoint #230

### LINKED ISSUE #230: Add usage statuses endpoint
see #170

joshuadkitenge:
## Description
- Add collection for usage statuses 
- Add get endpoint to get all usage statuses  
- Add validation for usages status when creating and updating an item 

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #230


test/e2e/conftest.py
```diff
@@ -21,6 +22,17 @@ def fixture_test_client() -> TestClient:
     return TestClient(app, headers={"Authorization": f"Bearer {VALID_ACCESS_TOKEN}"})
 
 
+@pytest.fixture(name="add_usage_statuses_to_database", autouse=True, scope="session")
+def fixture_add_usage_statuses_to_database():
+    """
+    Fixture to add usage statuses to the usage_statuses collection in the database before any tests in this file run
```
joelvdavies, line 28:
```suggestion
    Fixture to add usage statuses to the usage_statuses collection in the database before any tests run
```

test/e2e/test_item.py
```diff
@@ -935,11 +965,44 @@ def test_partial_update_item(test_client):
     assert item == {**ITEM_POST_EXPECTED, **item_patch, "catalogue_item_id": catalogue_item_id, "system_id": system_id}
 
 
+def test_partial_update_item_usage_status(test_client):
+    """
+    Test updating an item with a invalid usage status
+    """
```
joelvdavies, line 971: I dont think this test is needed `test_partial_update_item` already changes the usage status. Also this is checking the error currently the same as above.

README.md
```diff
@@ -214,16 +214,28 @@ To update the list of units, replacing all existing with the contents of the `./
...
 ```
 
+For usage statuses:
+```bash
+docker exec -i CONTAINER_ID mongoimport --username 'root' --password 'example' --authenticationDatabase=admin --db ims --collection usage_statuses --type=json --jsonArray --drop < ./data/usage_statuses.json
+```
+
 Replacing `CONTAINER_ID` with the actual container ID of the MongoDB instance.
 
 ### Using mock data for testing
```
joelvdavies, line 241: The `Generating new data` should also mention `units and usage statuses`.

README.md
```diff
@@ -201,10 +201,10 @@ payload, and it has not expired. This means that any microservice can be used to
 it meets the above criteria. The [LDAP-JWT Authentication Service](https://github.com/ral-facilities/ldap-jwt-auth) is
 a microservice that provides user authentication against an LDAP server and returns a JWT access token.
 
-### Adding units
+### Adding units and usage statuses
 
-Units should be added to the MongoDB database using `mongoimport` on the provided units file found at
-`/data/units.json`. If adding more units to this file, ensure the `_id` values are valid `ObjectId`'s.
+Units and usage statuses should be added to the MongoDB database using `mongoimport` on the provided units and usage statues file found at
```
joelvdavies, line 206:
```suggestion
Units and usage statuses should be added to the MongoDB database using `mongoimport` on the provided units and usage statues files found at
```

joelvdavies (commented):
A few minor things.  Should be merged with the front end changes.

As a general note this will require us to either write a migration, or delete the existing dev machine data. If we wanted I think this could actually be a nice migration, of course we will have the same for the property ids anyway so there is not much point.

README.md
```diff
@@ -201,10 +201,10 @@ payload, and it has not expired. This means that any microservice can be used to
 it meets the above criteria. The [LDAP-JWT Authentication Service](https://github.com/ral-facilities/ldap-jwt-auth) is
 a microservice that provides user authentication against an LDAP server and returns a JWT access token.
 
-### Adding units
+### Adding units and usage statuses
 
-Units should be added to the MongoDB database using `mongoimport` on the provided units file found at
-`/data/units.json`. If adding more units to this file, ensure the `_id` values are valid `ObjectId`'s.
+Units and usage statuses should be added to the MongoDB database using `mongoimport` on the provided units and usage statues files found at
```
joelvdavies, line 206:
```suggestion
Units and usage statuses should be added to the MongoDB database using `mongoimport` on the provided units and usage statuses files found at
```

joelvdavies:
@VKTB Should we be attempting to make a migration for this/the property ids? I could take a look if required. I suspect in this particular case its quite simple but the ids are more complicated. If we do I think a script inside the repo itself installed via the `pyproject.toml` would make sense in this particular case as it can use the same credentials and would be needed on the dev machine. Just depends how concerned we are with keeping the existing dev data.

If we do this however this PR should probably wait for that before being merged as it will be slightly easier to test migrations switching between the current dev branch and this.

inventory_management_system_api/services/item.py
```diff
@@ -74,6 +81,10 @@ def create(self, item: ItemPostRequestSchema) -> ItemOut:
         except InvalidObjectIdError as exc:
             raise DatabaseIntegrityError(str(exc)) from exc
 
+        usage_statuses = [usage_status.value for usage_status in self._usage_status_repository.list()]
+        if item.usage_status not in usage_statuses:
+            raise MissingRecordError(f"No usage status found with name: {item.usage_status}")
```
VKTB, line 86: Do we need to check this because we are not doing this for units and to me they are both of equal importance? So perhaps we need to do the same check for units if this is actually needed. But the more important question is whether the API should require for `usage_status_id` to be supplied by the client rather than name/value?

test/e2e/mock_schemas.py
```diff
@@ -132,3 +133,13 @@
...
+
+# Load in the usage statuses data
+with open("./data/usage_statuses.json", "r", encoding="utf-8") as file:
+    usage_statuses_data = json.load(file)
+    USAGE_STATUSES = [
+        {"_id": usage_status["_id"]["$oid"], "value": usage_status["value"]} for usage_status in usage_statuses_data
+    ]
+USAGE_STATUSES_EXPECTED = [
+    {"id": usage_status["_id"], "value": usage_status["value"]} for usage_status in USAGE_STATUSES
+]
```
VKTB, line 145: Do we have to load the usage statuses from the JSON file? Can't we just hardcode a couple of usage statuses (they can even be dummy ones)? I don't think we need to insert all the usage statuses for the purposes of testing anyway.

test/e2e/conftest.py
```diff
@@ -21,6 +22,17 @@ def fixture_test_client() -> TestClient:
...
 
+@pytest.fixture(name="add_usage_statuses_to_database", autouse=True, scope="session")
+def fixture_add_usage_statuses_to_database():
+    """
+    Fixture to add usage statuses to the usage_statuses collection in the database before any tests run
+    """
+    usage_statuses_collection = get_database().usage_statuses
+    usage_statuses_collection.delete_many({})
+    usage_statuses_collection.insert_many(USAGE_STATUSES)
+    yield
```
VKTB, line 33: Could this be moved to the `test_usage_status.py` `e2e` test module? Unless I am wrong that's the only module that requires it?

test/unit/services/test_item.py
```diff
@@ -98,6 +98,17 @@
...
+# Load in the usage statuses data
+with open("./data/usage_statuses.json", "r", encoding="utf-8") as file:
+    usage_statuses_data = json.load(file)
+    USAGE_STATUSES = [
+        {"_id": usage_status["_id"]["$oid"], "value": usage_status["value"]} for usage_status in usage_statuses_data
+    ]
+USAGE_STATUSES_LIST = [
+    UsageStatusOut(id=usage_status["_id"], value=usage_status["value"]) for usage_status in USAGE_STATUSES
+]
+
```
VKTB, line 111: Unit tests shouldn't interact with the file system as that makes tests less isolated and more susceptible to external factors. Could you please hardcode a couple of usage statuses and use those when mocking instead?

VKTB:
> @VKTB Should we be attempting to make a migration for this/the property ids? I could take a look if required. I suspect in this particular case its quite simple but the ids are more complicated. If we do I think a script inside the repo itself installed via the `pyproject.toml` would make sense in this particular case as it can use the same credentials using pymongo and it would be needed on the dev machine. Just depends how concerned we are with keeping the existing dev data.
> 
> If we do this however, this PR should probably wait for that before being merged as it will be slightly easier to test migrations switching between the current dev branch and this.

We can have a chat about this tomorrow.

inventory_management_system_api/services/item.py
joelvdavies: We did debate it ourselves, the reason units originally didn't have the check is because we had discussed it as being a placeholder to be edited later, but I do think it could be checked now although I prefer checking using the id rather than pulling the whole list and checking each value, so I feel they are one and the same issue (although we currently have less usage statuses). If we changed this to send back ids, we should do the same for units but as we don't have aggregations and its faster to just store the value, I don't see what the benefit would be to move to ids like that just yet.

test/e2e/mock_schemas.py
joelvdavies: We do the same for units, I agree it doesn't necessarily need all of them but its nice as a check that we haven't broken the syntax of the file. With full crud, this can be tested like the others. We also have the possibility of pre-populating the test data with the units, usage_statuses and predefined catalogue item property using the dev script in the CI if we wanted to (Which is what I have done on the front end api tests), although the expected values would still have to be loaded.

test/e2e/conftest.py
joelvdavies: Its also needed to do the item tests given that the check is in place for the create.

inventory_management_system_api/services/item.py
VKTB: > the API should require for `usage_status_id` to be supplied by the client rather than name/value?

By this, I meant that the API expects the `usage_status_id` to be supplied by the client in the item `POST` and `PATCH` requests but the API returns the actual usage status name/value (and potentially `usage_status_id`) in the response. This way it would check if the usage status exists using the ID as I also don't like retrieving the whole list and checking each one, and would store both the `usage_status_id` and usage name/value in the database for each item. Storing the IDs means that we won't have to do any ID back population when we get aggregations working nicely in future.

test/e2e/mock_schemas.py
VKTB: I see. Given that we are now doing checks in the API, it means that it now cares what usage status you give it so this functionality now makes sense to stay in the API and with this in mind, I believe that we should support full CRUD (maybe not update and delete to start with as it would require migrations). You mentioned that Joshua wanted to do the full CRUD for the property templates but I see this functionality here as more important. We should do the same for units.

test/e2e/conftest.py
VKTB: Ah yes, sorry, I didn't realise that.

test/e2e/mock_schemas.py
joshuadkitenge: > (maybe not update and delete to start with as it would require migrations)


I'm guessing the patch and delete should work like manufacturers. A user/admin should not be able to delete a usage status if the usage status is being used in an item, and the patch would work if we used IDs instead of the value, but if we use the value, it would be the same as the delete.

joelvdavies: > I see. Given that we are now doing checks in the API, it means that it now cares what usage status you give it so this functionality now makes sense to stay in the API and with this in mind, I believe that we should support full CRUD (maybe not update and delete to start with as it would require migrations). You mentioned that Joshua wanted to do the full CRUD for the property templates but I see this functionality here as more important. We should do the same for units.

Yeah that sounds sensible to me.



> > (maybe not update and delete to start with as it would require migrations)
> 
> I'm guessing the patch and delete should work like manufacturers. A user/admin should not be able to delete a usage status if the usage status is being used in an item, and the patch would work if we used IDs instead of the value, but if we use the value, it would be the same as the delete.

I think as there is no way to really find those items that use that usage status and it certainly doesn't feel very feasible to expect the user to change them if there are a lot, it does make more sense to me that the edit for an admin functionality like that would require a migration rather than a normal patch as Viktor suggested. Units certainly would pose another challenge in the way that they are stored if we were to try and look up whether they are used or not. Again I think a migration is the only way if they wanted to change the standard unit they use for something for example although I would hope no unit just ceases to exist.

VKTB:
As discussed, please implement the following in this PR:
- `POST` endpoint for creating usage statuses
- `GET` endpoint for getting usage status by ID
- Modify items to take in `usage_status_id`

test/unit/services/test_usage_status.py
```diff
@@ -0,0 +1,90 @@
...
+):
+    """Test getting a usage status with an non-existent ID"""
+    usage_status_id = str(ObjectId())
+    test_helpers.mock_get(usage_status_repository_mock, None)
+
+    # Mock `get` to return a usage status
+    retrieved_manufacturer = usage_status_service.get(usage_status_id)
+
+    assert retrieved_manufacturer is None
+    usage_status_repository_mock.get.assert_called_once_with(usage_status_id)
```
joshuadkitenge, line 79:
```suggestion
    # Mock `get` to return a usage status
    retrieved_usage_status = usage_status_service.get(usage_status_id)

    assert retrieved_usage_status is None
    usage_status_repository_mock.get.assert_called_once_with(usage_status_id)
```

inventory_management_system_api/services/item.py
```diff
@@ -74,6 +81,11 @@ def create(self, item: ItemPostRequestSchema) -> ItemOut:
         except InvalidObjectIdError as exc:
             raise DatabaseIntegrityError(str(exc)) from exc
 
+        usage_status_id = item.usage_status_id
+        usage_status = self._usage_status_repository.get(usage_status_id)
+        if not usage_status:
+            raise MissingRecordError(f"No usage status found with name: {usage_status_id}")
```
joshuadkitenge, line 87:
```suggestion
            raise MissingRecordError(f"No usage status found with ID: {usage_status_id}")
```

inventory_management_system_api/services/item.py
joelvdavies: This is now addressed in https://github.com/ral-facilities/inventory-management-system-api/pull/248

joelvdavies (changes requested):
Mostly minor comments to improve consistency and a couple missing tests. Worked well when I tested it.

inventory_management_system_api/repositories/usage_status.py
```diff
@@ -0,0 +1,94 @@
...
+            return UsageStatusOut(**usage_status)
+        return None
+
+    def _is_duplicate_usage_status(self, code: str, session: ClientSession = None) -> bool:
+        """
+        Check if usage status with the same name already exists in the usage statuses collection
+
+        :param code: The code of the usage status to check for duplicates.
+        :param session: PyMongo ClientSession to use for database operations
+        :return `True` if duplicate usage status, `False` otherwise
```
VKTB, line 91:
```suggestion
        :return: `True` if a duplicate usage status code is found, `False` otherwise
```

inventory_management_system_api/repositories/usage_status.py
```diff
@@ -0,0 +1,94 @@
...
+
+        :param usage_status_id: The ID of the usage status to retrieve.
+        :param session: PyMongo ClientSession to use for database operations
+        :return: The retrieved usage status, or `None` if not found.
+        """
+        usage_status_id = CustomObjectId(usage_status_id)
+        logger.info(
+            "Retrieving usage status with ID: %s from the database",
+            usage_status_id,
+        )
```
VKTB, line 79:
```suggestion
        logger.info("Retrieving usage status with ID: %s from the database", usage_status_id)
```

inventory_management_system_api/routers/v1/usage_status.py
```diff
@@ -0,0 +1,75 @@
...
+    # pylint: disable=missing-function-docstring
+    logger.info("Creating a new usage status")
+    logger.debug("Usage status data: %s", usage_status)
+
+    try:
+        usage_status = usage_status_service.create(usage_status)
+        return UsageStatusSchema(**usage_status.model_dump())
+
+    except DuplicateRecordError as exc:
+        message = "A usage status with the same name already exists"
```
joelvdavies, line 39:
```suggestion
        message = "A usage status with the same value already exists"
```
Just thought this is actually the value.

test/unit/repositories/test_usage_status.py
```diff
@@ -0,0 +1,177 @@
...
+
+from inventory_management_system_api.models.usage_status import (
+    UsageStatusIn,
+    UsageStatusOut,
+)
+
+
+def test_create(test_helpers, database_mock, usage_status_repository):
+    """
+     Test creating a usage status.
```
joelvdavies, line 24:
```suggestion
    Test creating a usage status.
```

scripts/generate_mock_data.py
```diff
@@ -383,32 +394,36 @@ def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
...
             timeout=10,
         )
         status_code = response.status_code
         index += 1
     return response.json()
 
 
-def create_manufacturer(manufacturer_data: dict) -> dict[str, str]:
-    return post_avoiding_duplicate_names(endpoint="/v1/manufacturers", json=manufacturer_data)
+def create_manufacturer(manufacturer_data: dict) -> str:
```
joelvdavies, line 405:
```suggestion
def create_manufacturer(manufacturer_data: dict) -> dict[str, Any]:
```

scripts/generate_mock_data.py
```diff
@@ -383,32 +394,36 @@ def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
...
+    return post_avoiding_duplicates(endpoint="/v1/manufacturers", field="name", json=manufacturer_data)
 
 
-def create_catalogue_category(category_data: dict) -> dict[str, str]:
-    return post_avoiding_duplicate_names(endpoint="/v1/catalogue-categories", json=category_data)
+def create_usage_status(usage_status_data: dict) -> str:
```
joelvdavies, line 409:
```suggestion
def create_usage_status(usage_status_data: dict) -> dict[str, Any]:

```

scripts/generate_mock_data.py
```diff
@@ -383,32 +394,36 @@ def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
...
+    return post_avoiding_duplicates(endpoint="/v1/usage-statuses", field="value", json=usage_status_data)
 
 
-def create_catalogue_item(item_data: dict) -> dict[str, str]:
-    return post_avoiding_duplicate_names(endpoint="/v1/catalogue-items", json=item_data)
+def create_catalogue_category(category_data: dict) -> str:
```
joelvdavies, line 413:
```suggestion
def create_catalogue_category(category_data: dict) -> dict[str, Any]:

```

scripts/generate_mock_data.py
```diff
@@ -383,32 +394,36 @@ def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
...
+    return post_avoiding_duplicates(endpoint="/v1/catalogue-categories", field="name", json=category_data)
 
 
-def create_system(system_data: dict) -> dict[str, str]:
-    return post_avoiding_duplicate_names(endpoint="/v1/systems", json=system_data)
+def create_catalogue_item(item_data: dict) -> str:
```
joelvdavies, line 417:
```suggestion
def create_catalogue_item(item_data: dict) -> dict[str, Any]:

```

scripts/generate_mock_data.py
```diff
@@ -383,32 +394,36 @@ def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
...
+    return post_avoiding_duplicates(endpoint="/v1/catalogue-items", field="name", json=item_data)
 
 
-def create_item(item_data: dict) -> dict[str, str]:
-    return post_avoiding_duplicate_names(endpoint="/v1/items", json=item_data)
+def create_system(system_data: dict) -> str:
```
joelvdavies, line 421:
```suggestion
def create_system(system_data: dict) -> dict[str, Any]:

```

scripts/generate_mock_data.py
```diff
@@ -371,7 +382,7 @@ def generate_random_system(parent_id):
     }
 
 
-def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
+def post_avoiding_duplicates(endpoint: str, field: str, json: dict) -> str:
```
joelvdavies, line 385:
```suggestion
def post_avoiding_duplicates(endpoint: str, field: str, json: dict) -> dict[str, Any]:

```

scripts/generate_mock_data.py
```diff
@@ -383,32 +394,36 @@ def post_avoiding_duplicate_names(endpoint: str, json: dict) -> str:
...
+    return post_avoiding_duplicates(endpoint="/v1/catalogue-items", field="name", json=item_data)
 
 
-def create_item(item_data: dict) -> dict[str, str]:
-    return post_avoiding_duplicate_names(endpoint="/v1/items", json=item_data)
+def create_system(system_data: dict) -> str:
+    return post_avoiding_duplicates(endpoint="/v1/systems", field="name", json=system_data)
+
+
+def create_item(item_data: dict) -> str:
```
joelvdavies, line 425:
```suggestion
def create_item(item_data: dict) -> dict[str, Any]:
```

joelvdavies (commented):
Just two more minor things, I also realised the types were slightly wrong in the generation script.

VKTB (approved):
Please squash and merge this when ready to be merged.

joelvdavies (approved):
Do not squash and merge until dev is migrated and functioning properly

test/e2e/test_catalogue_item.py
```diff
@@ -2117,13 +2129,14 @@ def test_partial_update_catalogue_item_properties_when_has_child_items(test_clie
     }
     response = test_client.post("/v1/catalogue-items", json=catalogue_item_post)
     catalogue_item_id = response.json()["id"]
-
+    # pylint: disable=duplicate-code
```
VKTB, line 2132:
```suggestion
```

test/e2e/test_item.py
```diff
@@ -101,18 +101,22 @@ def test_create_item(test_client):
...
         "manufacturer_id": manufacturer_id,
         "properties": add_ids_to_properties(
-            catalogue_category["catalogue_item_properties"], CATALOGUE_ITEM_POST_A["properties"]
+            catalogue_category["catalogue_item_properties"],
+            CATALOGUE_ITEM_POST_A["properties"],
         ),
     }
     response = test_client.post("/v1/catalogue-items", json=catalogue_item_post)
     catalogue_item_id = response.json()["id"]
-    # pylint: enable=duplicate-code
```
VKTB, line 109: Should this be removed?

inventory_management_system_api/models/usage_status.py
```diff
@@ -0,0 +1,34 @@
...
+from inventory_management_system_api.models.custom_object_id_data_types import StringObjectIdField
+from inventory_management_system_api.models.mixins import CreatedModifiedTimeInMixin, CreatedModifiedTimeOutMixin
+
+
+class UsageStatusBase(BaseModel):
+    """
+    Base database model for a UsageStatus
+    """
+
+    value: str = Field(description="Value of the Usage status")
```
VKTB, line 16:
```suggestion
    value: str
```

inventory_management_system_api/models/usage_status.py
```diff
@@ -0,0 +1,34 @@
...
+    # Used for uniqueness checks (sanitised value)
+    code: str = Field(description="Code of the Usage status")
```
VKTB, line 18:
```suggestion
    code: str
```

test/e2e/test_item.py
joshuadkitenge: Yes, but I forgot to remove Line 114

VKTB (approved):

joelvdavies (approved):

PR #234, opened 2024-04-11
Add endpoint for predefined catalogue item properties #233

### LINKED ISSUE #233: Add endpoint for GET list of predefined catalogue item properties

joshuadkitenge:
## Description
- Add collection for catalogue item property templates 
- Add get endpoint for getting all catalogue item property templates

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes #233

### DIFF
--- a/inventory_management_system_api/models/catalogue_item_property_template.py
+++ b/inventory_management_system_api/models/catalogue_item_property_template.py
@@ -2,26 +2,17 @@
 Module for defining the database models for representing a Catalogue item property template
 """
 
-from typing import Optional
+from pydantic import ConfigDict, Field
 
-from pydantic import BaseModel, ConfigDict, Field
-
-from inventory_management_system_api.models.catalogue_category import AllowedValues
+from inventory_management_system_api.models.catalogue_category import CatalogueItemProperty
 from inventory_management_system_api.models.custom_object_id_data_types import StringObjectIdField
 
-# pylint: disable=duplicate-code
 
-
-class CatalogueItemPropertyTemplateOut(BaseModel):
+class CatalogueItemPropertyTemplateOut(CatalogueItemProperty):
     """
     Model representing a catalogue item property template.
     """
 
     id: StringObjectIdField = Field(alias="_id")
-    name: str
-    type: str
-    unit: Optional[str] = None
-    mandatory: bool
-    allowed_values: Optional[AllowedValues] = None
 
     model_config = ConfigDict(populate_by_name=True)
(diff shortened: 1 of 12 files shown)


VKTB:
Quick question: How are these templates going to be added?

inventory_management_system_api/models/catalogue_item_property_template.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+
+from typing import Optional
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from inventory_management_system_api.models.catalogue_category import AllowedValues
+from inventory_management_system_api.models.custom_object_id_data_types import StringObjectIdField
+
+# pylint: disable=duplicate-code
```
joelvdavies, line 12: Could you enable this at the end of the file again, its easy to forget about if the file is added to.

joshuadkitenge:
> Quick question: How are new templates going to be added?

It is going to be imported like the units. However I have time to implement a full CRUD (frontend and backend) for this before MVP. I can start the implement of this after I have done the UI changes for the migration of properties which I'm starting today.

inventory_management_system_api/schemas/catalogue_item_property_template.py
```diff
@@ -0,0 +1,28 @@
...
+
+from typing import Optional
+
+from pydantic import BaseModel, Field
+
+
+from inventory_management_system_api.schemas.catalogue_category import AllowedValuesSchema, CatalogueItemPropertyType
+
+
+# pylint: disable=duplicate-code
```
joelvdavies, line 13: Again could you enable this at the end of the file please

joelvdavies (commented):
Two minor things. Also the readme/test data dump needs updating again. I made #236, to try and make this all simpler in the future, so I could also update it there.

inventory_management_system_api/models/catalogue_item_property_template.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class CatalogueItemPropertyTemplateOut(BaseModel):
+    """
+    Model representing a catalogue item property template.
+    """
+
+    id: StringObjectIdField = Field(alias="_id")
```
joelvdavies, line 20: I see on the front end you just inherit and add the id. Cant you do the same with the model and schema here? The schema in particular also has some validation that would be needed if doing a create endpoint. Of course Viktor's PR will be adding IDs as well so perhaps its even simpler.

VKTB:
> It is going to be imported like the units. However I have time to implement a full CRUD (frontend and backend) for this before MVP. I can start the implement of this after I have done the UI changes for the migration of properties which I'm starting today.

If this can be delivered as it is then I would rather address the full CRUD after the MVP. I think we discussed that this and other related functionality make sense to be in a separate service since it is only of use to the frontend but not the IMS API itself.

joshuadkitenge:
#245

PR #242, opened 2024-04-16
Add session parameter to all repo methods #239

### LINKED ISSUE #239: Implement way for using session inside repo methods
We need a way of using MongoDB sessions inside the repo layer of methods for the purpose of applying transactions for migrations.

Looking at
https://stackoverflow.com/questions/74073516/do-i-really-have-to-pass-the-session-as-a-param-when-doing-mongo-node-transactio
and
https://stackoverflow.com/questions/55980425/why-are-many-sessions-faster-in-mongodb-than-one

It appears adding a session parameter to the repo methods is the most appropriate.

joelvdavies:
## Description
I thought this may be the best approach for #239, but happy to discuss. With this we can reuse any and all repo methods, but pass in the session after the transaction has started. Also has potential of use in consistent sessions (not that we need them), as get requests that happen after edit can use the same session in the update methods for example.

Note: Also replaces a few `nonexistent` -> `non-existent`, and fixes a mistake in move check tests (it was still mocking as though it was for breadcrumbs and didn't check the aggregate function was called)

Would also want these changes for #231, #234.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #239

### DIFF
--- a/inventory_management_system_api/repositories/catalogue_category.py
+++ b/inventory_management_system_api/repositories/catalogue_category.py
@@ -6,6 +6,7 @@ import logging
 from typing import List, Optional
 
 from fastapi import Depends
+from pymongo.client_session import ClientSession
 from pymongo.collection import Collection
 from pymongo.database import Database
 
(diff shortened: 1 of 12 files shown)


VKTB (approved):
LGTM

PR #248, opened 2024-04-19
integrate unit id in catalogue category , item and catalogue items  #243

joshuadkitenge:
## Description
see #243

## Testing instructions
Add a set up instructions describing how the reviewer should test the code

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #243

PR #250, opened 2024-04-19
Migration script for renaming catalogue_item_properties to properties #249

### LINKED ISSUE #249: Create migration script for populating property ids
Will only be needed for dev machine data, but also serves as an investigation into doing these in the future once in production.

joelvdavies:
## Description

Adds a command to do migrations. This does not necessarily need to be merged, the `add_property_ids` script is only required for dev, so simply pointing to this PR's docker image temporarily should be sufficient.

To use
```
ims-migrate add_property_ids
```
Or to migrate backwards
```
ims-migrate add_property_ids --backward
```
This can be done within docker using 
```
docker exec -it inventory_management_system_api_container ims-migrate add_property_ids
```
Transactions are used here in case of failure mid migration (this should also in theory lock the database for writes while the migration is happening).

If running on dev first take a `mongodump` before attempting the migration.

This pr also has a script for adding usage_status_ids as in #231 and unit_ids as in #248.

## General thoughts on migrations
This PR follows a similar method to beanie which installs a migrate script with the package. This is also similar to alembic which uses the same client auth as SQLAlchemy to run its migration. In general it appears migrations tend to be run on application start, or as part of the CI build process, or via another container entirely (see https://medium.com/codex/database-migration-when-your-service-is-running-in-kubernetes-abbe9697421d for potential methods).

In our case for production we likely want to shut off the instance, and then run the migrations before starting up again. This is tricky using the docker container itself as the container needs to be running in order for the script to be usable here. As a result running it locally may be the better option here although it will be a bit more involved for whoever does the update. For the dev machine there is the added complexity of the docker compose networking, which will prevent outside access. In these cases migrations appear to be done inside one of the containers itself as is done here.

For most migrations we will likely need the model schemas inside the migration scripts. This ensures the order of parameters is kept consistent e.g. when adding usage_status_id, and also provides an additional layer of validation prior to changes which would allow a fallback in the event of finding something in a format the script doesn't expect. This is where beanie can help, although it is also possible to recreate some of the behaviour ourselves.
 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #249


VKTB:
I tested this and it works nicely, Thank you for looking into it!

joelvdavies:
Closing as all of the migrations this script was made for have been completed.

joelvdavies:
Re-opening for #293

joelvdavies:
Closing again as migration for #294 is complete

PR #273, opened 2024-05-17
Add check for duplicate values in an allowed_values list #272

### LINKED ISSUE #272: Add check for duplicate values in an allowed values list
Currently not checked.

joelvdavies:
## Description
Checks for catalogue category post and patch, and the new properties post and patch. String values are converted to lower case first like the front end check so two values with different capitalisation are treated as duplicates.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #272

### DIFF
--- a/inventory_management_system_api/schemas/catalogue_category.py
+++ b/inventory_management_system_api/schemas/catalogue_category.py
@@ -114,8 +114,11 @@ class CatalogueCategoryPostRequestPropertySchema(BaseModel):
                 )
             # Check the type of allowed_values being used and validate them appropriately
             if isinstance(allowed_values, AllowedValuesListSchema):
-                # List type should have all values the same type
+                # List type should have all values the same type and no duplicate values
+                seen_values = set()
+
                 for allowed_value in allowed_values.values:
+                    # Ensure the value is the correct type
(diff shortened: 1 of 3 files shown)


inventory_management_system_api/schemas/catalogue_category.py
```diff
@@ -124,6 +127,16 @@ def check_valid_allowed_values(
                             "itself"
                         )
 
+                    # Ensure the value isn't duplicated
+                    seen_value = (
+                        allowed_value
+                        if catalogue_item_property_data["type"] != CatalogueItemPropertyType.STRING
+                        else allowed_value.lower()
+                    )
+                    if seen_value in seen_values:
```
VKTB, line 136: Could you please update the docstring of this method to mention this check?

VKTB (commented):
Just one thing related to docstrings, works well otherwise.

inventory_management_system_api/schemas/catalogue_category.py
joelvdavies: I moved it to the :raises section as I thought it might be clearer to list them as bullet points.

PR #284, opened 2024-06-06
Handling sensitive config values

VKTB:
## Description
This PR changes the types of the `DatabaseConfig` Pydantic class fields from `str` to `SecretStr`. This means that they will be serialised to ``**********`` when accessing their values, or using `.model_dump()`, `.dict()`, or `.json()` (more info [here](https://docs.pydantic.dev/latest/examples/secrets/)). `get_secret_value()` should be used to access the secret value.

It also sets `hide_input_in_errors` on the models to `True` so that the input values can be hidden when `ValidationError` is raised during the validation.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Ensure that input values are not part of the error messages and logs (A validation error can be triggered by supplying a wrong type of value for a config or not supplying a value for a mandatory config)

### DIFF
--- a/inventory_management_system_api/main.py
+++ b/inventory_management_system_api/main.py
@@ -5,7 +5,7 @@ Main module contains the API entrypoint.
 import logging
 
 from fastapi import Depends, FastAPI, Request, status
-from fastapi.encoders import jsonable_encoder
+from fastapi.exception_handlers import request_validation_exception_handler
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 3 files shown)


inventory_management_system_api/main.py
```diff
@@ -44,20 +44,23 @@ async def custom_general_exception_handler(_: Request, exc: Exception) -> JSONRe
...
     :param exc: The exception object representing the validation error.
     :return: A JSON response with validation error details.
     """
-    logger.exception(exc)
-    return await request_validation_exception_handler(request, exc)
+    # Exclude input as it may contain sensitive information which we do not want to log or return back to user
+    detail = {"detail": jsonable_encoder(exc.errors(), exclude={"input"})}
+    # The traceback here contains the exception message which includes the input hence it is been excluded.
+    logger.exception(detail, exc_info=False)
+    return JSONResponse(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, content=detail)
```
VKTB, line 63: I added this in but I wanted to check if we actually need it? We are not working with any sensitive data at an endpoint level like we do with usernames and passwords in the Auth component. The only case I can think it would be useful is if a user accidentally adds some sensitive data in their request, a validation request is raised, and that sensitive data ends up in the logs.

VKTB: > if a user accidently pastes a password into the name field, its not going to complain and will log it anyway for the post requests and then it will be in the database as well

That's true. Okay I will just revert these changes then, thanks.

joelvdavies (approved):

PR #294, opened 2024-06-14
Rename catalogue_item_properties to properties #293

### LINKED ISSUE #293: Change name of catalogue_item_properites in catalogue categories
Will require front end changes and a migration for dev.

joelvdavies:
## Description
Renames
- catalogue_item_properties -> properties
- catalogue_item_property -> property
- CatalogueItemProperty -> CatalogueCategoryProperty

Where necessary have replaced the resulting `property` variable name with either `property_in` for database models, or `category_property` for schemas.

See https://github.com/ral-facilities/inventory-management-system/pull/662 for front end.

Should not be merged until dev is migrated.

### Note
Could we also simplify some schema names by removing the `Request` in them? The method type already implies its a request to me. In fact in the front end we don't even have the schema e.g. `CatalogueItemPostRequestSchema` could be `CatalogueItemPost` which is imported from schemas.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #293


inventory_management_system_api/core/exceptions.py
```diff
@@ -21,21 +21,21 @@ class NonLeafCategoryError(Exception):
...
-    The type of the provided value does not match the expected type of the catalogue item property.
+    The type of the provided value does not match the expected type of the property.
     """
 
 
-class MissingMandatoryCatalogueItemProperty(Exception):
+class MissingMandatoryProperty(Exception):
     """
-    A mandatory catalogue item property is missing when a catalogue item is attempted to be created.
+    A mandatory property is missing when a catalogue item is attempted to be created.
```
VKTB, line 38: I think this one is also raised when an item is attempted to be created so should we update the docstring to mention item?

inventory_management_system_api/models/catalogue_category.py
```diff
@@ -24,9 +24,9 @@ class AllowedValuesList(BaseModel):
 AllowedValues = Annotated[AllowedValuesList, Field(discriminator="type")]
 
 
-class CatalogueItemPropertyBase(BaseModel):
+class CategoryPropertyBase(BaseModel):
```
VKTB, line 27: Same question - Just `Category` or should it be `CatalogueCategory`?

inventory_management_system_api/models/catalogue_item.py
VKTB, line 1: I think what we can do here instead is rename these to `CatalogueItemPropertyIn` and `CatalogueItemPropertyOut` and then in `/models/item.py`, we define `ItemPropertyIn` and `ItemPropertyOut` that inherit from these two. This way we have classes for each one. Let me know your thoughts.

VKTB (commented):
> Could we also simplify some schema names by removing the Request in them? The method type already implies its a request to me. In fact in the front end we don't even have the schema e.g. CatalogueItemPostRequestSchema could be CatalogueItemPost which is imported from schemas.

Possibly but we need to make sure that there are no conflicts. It would be an issue if we end up with conflicting class names and having to import this in the same module.

Some initial questions below

inventory_management_system_api/core/exceptions.py
```diff
@@ -21,21 +21,21 @@ class NonLeafCategoryError(Exception):
     """
 
 
-class DuplicateCatalogueItemPropertyNameError(Exception):
+class DuplicateCategoryPropertyNameError(Exception):
```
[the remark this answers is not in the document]
joelvdavies: Yeah wasn't sure either, I think I will just go with CatalogueCategory to be clear.

joshuadkitenge (commented):
Works well with the front end

inventory_management_system_api/core/exceptions.py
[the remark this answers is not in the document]
joelvdavies: Although looking at it again, we also have `LeafCategoryError`, and DuplicateCatalogueCategoryPropertyNameError is quite long, do you prefer one over the other?

inventory_management_system_api/models/catalogue_item.py
joelvdavies: I like that in principle, if we do this here, we should also do the same for schemas. In that case there is also a potential issue where we have generic functions that apply for both, then they would be taking in the base class type which implies its CatalogueItem related when in fact Item ones will work too. Although I guess we could also make it explicit with CatalogueItem | Item.

joelvdavies:
> Possibly but we need to make sure that there are no conflicts. It would be an issue if we end up with conflicting class names and having to import this in the same module.
> 
Yeah good point, having Schema would resolve that part. I know this also effects some things Joshua is doing on the front end, as there Schema is now being used for form schemas whereas the api schemas are just in a types file without any mention of schema. I think its worth coordinating between us but its a post mvp thing.

inventory_management_system_api/models/catalogue_item.py
VKTB: As discussed, we should leave this for the time being.

joelvdavies: We discussed and decided to ignore for now as it creates other potential issues when implementing.

joelvdavies:
@VKTB I had a look at the other service's, I think the use of properties don't need renaming as they are specific to the entity involved e.g. CatalogueCategoryService only deals with catalogue category properties. So I think it was just the property service that needed the renaming as it deals with all of them unless you spot anything else.

inventory_management_system_api/routers/v1/catalogue_category.py
```diff
@@ -263,39 +263,39 @@ def create_catalogue_item_property(
...
+    summary="Update property at the catalogue category level",
+    response_description="The updated property as defined at the catalogue category level",
 )
-def partial_update_catalogue_item_property(
-    catalogue_item_property: CatalogueItemPropertyPatchRequestSchema,
+def partial_update_property(
+    category_property: CategoryPropertyPatchRequestSchema,
     catalogue_category_id: str = Path(description="The ID of the catalogue category containing the property to patch"),
-    catalogue_item_property_id: str = Path(description="The ID of the catalogue item property patch"),
+    property_id: str = Path(description="The ID of the property patch"),
```
VKTB, line 273:
```suggestion
    property_id: str = Path(description="The ID of the property to patch"),
```

inventory_management_system_api/services/catalogue_category.py
```diff
@@ -161,43 +159,39 @@ def update(
...
 
-    def _add_catalogue_item_property_unit_values(
+    def _add_property_unit_values(
         self,
-        catalogue_item_properties: List[CatalogueCategoryPostRequestPropertySchema],
+        properties: List[CatalogueCategoryPostRequestPropertySchema],
     ) -> List[dict[str, Any]]:
         """
-        Adds the unit values to the catalogue item properties based on the provided unit IDs.
+        Adds the unit values to the  based on the provided unit IDs.
```
VKTB, line 179:
```suggestion
        Adds the unit values to the properties based on the provided unit IDs.
```

inventory_management_system_api/services/catalogue_category.py
```diff
@@ -161,43 +159,39 @@ def update(
...
 
-        :param catalogue_item_properties: List of catalogue item properties to which unit values will be added.
-        :return: List of catalogue item properties with unit values added.
+        :param properties: List of  to which unit values will be added.
```
VKTB, line 181:
```suggestion
        :param properties: List of properties to which unit values will be added.
```

inventory_management_system_api/services/catalogue_category.py
```diff
@@ -161,43 +159,39 @@ def update(
...
+        :return: List of  with unit values added.
```
VKTB, line 182:
```suggestion
        :return: List of properties with unit values added.
```

inventory_management_system_api/services/catalogue_category.py
```diff
@@ -161,43 +159,39 @@ def update(
...
         :raises MissingRecordError: If a unit with the specified ID is not found.
         """
-        logger.info("Adding unit values to the catalogue item properties")
-        catalogue_item_properties_with_units = []
-        for catalogue_item_property in catalogue_item_properties:
-            if catalogue_item_property.unit_id is not None:
-                unit = self._unit_repository.get(catalogue_item_property.unit_id)
+        logger.info("Adding unit values to the ")
```
VKTB, line 185:
```suggestion
        logger.info("Adding unit values to the properties")
```

inventory_management_system_api/services/catalogue_category_property.py
```diff
@@ -181,71 +177,71 @@ def _check_valid_allowed_values_update(
     def update(
         self,
         catalogue_category_id: str,
-        catalogue_item_property_id: str,
-        catalogue_item_property: CatalogueItemPropertyPatchRequestSchema,
-    ) -> CatalogueItemPropertyOut:
+        catalogue_category_property_id: str,
+        category_property: CategoryPropertyPatchRequestSchema,
```
VKTB, line 181:
```suggestion
        catalogue_category_property: CategoryPropertyPatchRequestSchema,
```

inventory_management_system_api/services/catalogue_category_property.py
```diff
@@ -181,71 +177,71 @@ def _check_valid_allowed_values_update(
...
 
         Property changes will be propagated down through the catalogue items and items when required where there are
         children
 
         :param catalogue_category_id: The ID of the catalogue category to update
-        :param catalogue_item_property_id: The ID of the catalogue item property within the category to update
-        :param catalogue_item_property: The property values to update
-        :raises MissingRecordError: If the catalogue category doesn't exist, or the catalogue item property doesn't
+        :param catalogue_category_property_id: The ID of the property within the category to update
+        :param property: The property values to update
```
VKTB, line 191:
```suggestion
        :param catalogue_category_property: The property values to update
```

inventory_management_system_api/routers/v1/catalogue_category.py
```diff
@@ -224,27 +224,27 @@ def delete_catalogue_category(
...
-    summary="Create a new catalogue item property at the catalogue category level",
-    response_description="The created catalogue item property as defined at the catalogue category level",
+    summary="Create a new property at the catalogue category level",
+    response_description="The created property as defined at the catalogue category level",
     status_code=status.HTTP_201_CREATED,
 )
-def create_catalogue_item_property(
-    catalogue_item_property: CatalogueItemPropertyPostRequestSchema,
+def create_property(
+    category_property: CategoryPropertyPostRequestSchema,
```
VKTB, line 232:
```suggestion
    catalogue_category_property: CategoryPropertyPostRequestSchema,
```

inventory_management_system_api/routers/v1/catalogue_category.py
```diff
@@ -263,39 +263,39 @@ def create_catalogue_item_property(
...
-    summary="Update catalogue item property at the catalogue category level",
-    response_description="The updated catalogue item property as defined at the catalogue category level",
+    path="/{catalogue_category_id}/properties/{property_id}",
+    summary="Update property at the catalogue category level",
+    response_description="The updated property as defined at the catalogue category level",
 )
-def partial_update_catalogue_item_property(
-    catalogue_item_property: CatalogueItemPropertyPatchRequestSchema,
+def partial_update_property(
+    category_property: CategoryPropertyPatchRequestSchema,
```
VKTB, line 271:
```suggestion
    catalogue_category_property: CategoryPropertyPatchRequestSchema,
```

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -57,21 +57,21 @@ def create(self, catalogue_item: CatalogueItemPostRequestSchema) -> CatalogueIte
         Create a new catalogue item.
 
         The method checks if the catalogue category exists in the database and raises a `MissingRecordError` if it does
-        not. It also checks if the category is not a leaf category and raises a `NonLeafCategoryError` if it is. It then
-        processes the catalogue item properties.
+        not. It also checks if the category is not a leaf category and raises a `NonLeafCatalogueCategoryError` if it
+        is. It then processes the catalogue item properties.
```
VKTB, line 61:
```suggestion
        is. It then processes the properties.
```

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -131,7 +131,8 @@ def update(self, catalogue_item_id: str, catalogue_item: CatalogueItemPatchReque
         The method checks if the catalogue item exists in the database and raises a `MissingRecordError` if it does
         not. If the catalogue category ID is being updated, it checks if catalogue category with such ID exists and
         raises a MissingRecordError` if it does not. It also checks if the category is not a leaf category and raises a
-        `NonLeafCategoryError` if it is. If the catalogue item properties are being updated, it also processes them.
+        `NonLeafCatalogueCategoryError` if it is. If the catalogue item properties are being updated, it also processes
+        them.
```
VKTB, line 135:
```suggestion
        `NonLeafCatalogueCategoryError` if it is. If the properties are being updated, it also processes them.
```

inventory_management_system_api/services/item.py
```diff
@@ -132,7 +132,7 @@ def update(self, item_id: str, item: ItemPatchRequestSchema) -> ItemOut:
         not. If the system ID is being updated, it checks if the system ID with such ID exists and raises
         a `MissingRecordError` if it does not. It raises a `ChildElementsExistError` if a catalogue item
         ID is supplied. When updating properties, existing properties must all be supplied, or they will
-        be overwritten by the catalogue item properties.
+        be overwritten by the .
```
VKTB, line 135:
```
        be overwritten by the properties.
```

inventory_management_system_api/services/utils.py
```diff
@@ -35,80 +35,72 @@ def generate_code(name: str, entity_type: str) -> str:
...
+    properties: list[CatalogueCategoryPostRequestPropertySchema | CatalogueCategoryPropertyOut],
 ) -> None:
     """
-    Go through a list of catalogue item properties to check for any duplicate names during creation of a catalogue
+    Go through a list of  to check for any duplicate names during creation of a catalogue
     category or an addition of a property later on.
 
-    :param catalogue_item_properties: The supplied catalogue item properties
-    :raises DuplicateCatalogueItemPropertyName: If a duplicate catalogue item property name is found.
+    :param properties: The supplied
```
VKTB, line 45:
```
    Go through a list of properties to check for any duplicate names during creation of a catalogue category or an
    addition of a property later on.

    :param properties: The supplied properties
```

inventory_management_system_api/services/utils.py
```diff
@@ -35,80 +35,72 @@ def generate_code(name: str, entity_type: str) -> str:
...
+def process_properties(
+    defined_properties: List[CatalogueCategoryPropertyOut],
     supplied_properties: List[PropertyPostRequestSchema],
 ) -> List[Dict]:
     """
-    Process and validate supplied catalogue item properties based on defined catalogue item properties. It checks
-    for missing mandatory catalogue item properties, filters the matching catalogue item properties, adds the property
-    units, and finally validates the property values.
+    Process and validate supplied properties based on the defined properties. It checks for missing mandatory,
+    filters the matching , adds the property units, and finally validates the property values.
```
VKTB, line 63:
```
    Process and validate supplied properties based on the defined properties. It checks for missing mandatory, filters
    the matching properties, adds the property units, and finally validates the property values.
```

inventory_management_system_api/services/utils.py
```diff
@@ -35,80 +35,72 @@ def generate_code(name: str, entity_type: str) -> str:
...
+    # Some mandatory  may not have been supplied
+    _check_missing_mandatory_properties(defined_properties_dict, supplied_properties_dict)
+    # Some non-mandatory  may not have been supplied
+    supplied_properties_dict = _merge_non_mandatory_properties(defined_properties_dict, supplied_properties_dict)
+    # Supplied  do not have names as we can't trust they would be correct
+    _add_property_names(defined_properties_dict, supplied_properties_dict)
+    # Supplied  do not have units as we can't trust they would be correct
+    _add_property_units(defined_properties_dict, supplied_properties_dict)
+    # The values of the supplied  may not be of the expected types
+    _validate_property_values(defined_properties_dict, supplied_properties_dict)
```
VKTB, line 84:
```python
    # Some mandatory properties may not have been supplied
    _check_missing_mandatory_properties(defined_properties_dict, supplied_properties_dict)
    # Some non-mandatory properties may not have been supplied
    supplied_properties_dict = _merge_non_mandatory_properties(defined_properties_dict, supplied_properties_dict)
    # Supplied properties do not have names as we can't trust they would be correct
    _add_property_names(defined_properties_dict, supplied_properties_dict)
    # Supplied properties do not have units as we can't trust they would be correct
    _add_property_units(defined_properties_dict, supplied_properties_dict)
    # The values of the supplied properties may not be of the expected types
    _validate_property_values(defined_properties_dict, supplied_properties_dict)
```

inventory_management_system_api/services/utils.py
```diff
@@ -35,80 +35,72 @@ def generate_code(name: str, entity_type: str) -> str:
...
-    :return: A dictionary where the keys are the catalogue item property IDs and the values are the catalogue item
+    :param properties: The list of property objects.
+    :return: A dictionary where the keys are the property IDs and the values are the catalogue item
         property dictionaries.
     """
-    return {
-        catalogue_item_property.id: catalogue_item_property.model_dump()
-        for catalogue_item_property in catalogue_item_properties
-    }
+    return {property.id: property.model_dump() for property in properties}
```
VKTB, line 100: Shadows the built-in `property` name.
```suggestion
    return {prop.id: prop.model_dump() for prop in properties}
```

inventory_management_system_api/services/utils.py
```diff
@@ -137,25 +129,25 @@ def _add_catalogue_item_property_units(
     The supplied properties only contain an ID and value so the units from the defined properties in the database
     are added to the supplied properties. This means that this method modifies the `supplied_properties` dictionary.
 
-    :param defined_properties: The defined catalogue item properties stored as part of the catalogue category in the
+    :param defined_properties: The defined properties stored as part of the catalogue category in the
         database.
-    :param supplied_properties: The supplied catalogue item properties.
+    :param supplied_properties: The supplied .
```
VKTB, line 134:
```suggestion
    :param supplied_properties: The supplied properties.
```

VKTB (changes requested):

inventory_management_system_api/routers/v1/catalogue_category.py
```diff
@@ -263,39 +263,39 @@ def create_catalogue_item_property(
...
     # pylint: disable=missing-function-docstring
     logger.info(
-        "Partially updating catalogue category with ID %s's catalogue item property with ID: %s",
+        "Partially updating catalogue category with ID %s's property with ID: %s",
         catalogue_category_id,
-        catalogue_item_property_id,
+        property_id,
     )
-    logger.debug("Catalogue item property data: %s", catalogue_item_property)
+    logger.debug("Catalogue item property data: %s", category_property)
```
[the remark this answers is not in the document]
joelvdavies: 🤦‍♂️ I forgot to search for that particular case combination

VKTB:
@joelvdavies A lot of the requested changes are related to `catalogue item property` / `catalogue item properties` being replaced with an empty space instead of `property` / `properties`. This is the case in the tests but I didn't go and suggest changes as it would take me quite some time.

joelvdavies:
@VKTB Sorry about that, I thought I had only done it to a few files but I didn't realise I had done it everywhere. I have gone through the tests searching for double spaces for ` .` and fixed the rest I found.

joshuadkitenge (approved):

PR #299, opened 2024-06-19
Initial refactor of SystemRepo & Service unit tests #302

### LINKED ISSUE #302: Refactor systems tests
As part of #90. Using this as a first draft of proposed test changes that can be carried through the rest.

joelvdavies:
## Description
Initial refactor example for system repo unit tests. Code is not actually much shorter for these tests, but I think it is more readable and hopefully easier to maintain as the core test functionality is in one place (there are also some additional asserts that were missing before). 

SystemService are actually slightly longer now & the e2e tests are a bit shorter.

Suggestions welcome.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #302


test/unit/repositories/test_system.py
```diff
@@ -47,756 +47,764 @@
...
+        self.test_helpers.mock_find_one(
+            self.systems_collection, {**self._system_in.model_dump(), "_id": inserted_system_id}
+        )
+
+    def call_create(self):
+        """Calls the SystemRepo `create` method with the appropriate data from a prior call to `mock_create`"""
+
+        self._created_system = self.system_repository.create(self._system_in, session=self.mock_session)
+
+    def call_create_expecting_error(self, error_type: type):
```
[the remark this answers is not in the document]
joelvdavies: We could do, but I am not sure if its in the spirit of this method https://medium.com/@bas.dijkstra54/making-test-frameworks-readable-the-domain-specific-language-c154c9a9abcb. Abstracting further would effectively mean a single function for each test with a bunch of parameters when really they are supposed to be readable through the function names in logical steps. Where right now each are in the format mock, call, assert.

test/unit/repositories/test_system.py
```diff
@@ -47,756 +47,764 @@
...
-    database_mock.systems.insert_one.assert_called_once_with(
-        {**system_info, "parent_id": CustomObjectId(system_out.parent_id)}, session=session
-    )
-    database_mock.systems.find_one.assert_has_calls(
-        [
-            call({"_id": CustomObjectId(system_out.parent_id)}, session=session),
+class SystemRepoDSL:
+    """Base class for SystemRepo unit tests"""
+
+    test_helpers: RepositoryTestHelpers
```
joelvdavies, line 53: It might be worth replacing RepositoryTestHelpers with a base class that the same methods, it would save adding fixtures in a setup function?

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
+            else None
+        )
+        self.test_helpers.mock_get(self.mock_system_repository, self._stored_system)
+
+        # Patch schema
+        self._system_patch = SystemPatchSchema(**system_patch_data)
+
+        # Updated system
+        self._expected_system_out = MagicMock()
+        self.test_helpers.mock_update(self.mock_system_repository, self._expected_system_out)
```
joelvdavies, line 263: I am not sure this is necessary - I could equally use self.mock_system_repository.get.return_value in the assert later instead of storing this which would make it a bit shorter, although similar to a lot of things in this test file, right now the service layer doesn't do much, later we may need to mock like this with realistic data if there was processing after it to test. This may also be more readable here.

test/e2e/test_system.py
```diff
@@ -4,603 +4,579 @@
...
+        self.get_system_breadcrumbs("invalid_id")
+        self.check_get_breadcrumbs_failed_with_message(404, "System not found")
 
-    # Child - post twice as will have the same name
-    test_client.post("/v1/systems", json={**SYSTEM_POST_B, "parent_id": parent_system["id"]})
-    response = test_client.post("/v1/systems", json={**SYSTEM_POST_B, "parent_id": parent_system["id"]})
 
-    assert response.status_code == 409
-    assert response.json()["detail"] == "A System with the same name already exists within the same parent System"
+class ListDSL(GetBreadcrumbsDSL):
```
joelvdavies, line 284: I think its also worth point out I struggled to come up with a consistent naming system throughout these tests, right now its using `_system` or `_systems` in the method names to distinguish between these and later ones for items where we may be able to reuse logic but its not clear to me until we get that far so could be a case of renaming these then. 

Same for the class and method names - should they be the request type or not? Here I have used the request type for methods, but Get and List inside the class names simply because I didn't see the worth in putting System/Systems after each, but again it may be obvious later which is best.

test/e2e/test_system.py
```diff
@@ -4,603 +4,579 @@
...
-    assert response.json() == {"trail": [[system["id"], system["name"]] for system in systems], "full_trail": True}
+    def test_partial_update_name_to_duplicate(self):
+        """Test updating the name of a System to conflict with a pre-existing one"""
 
+        self.post_system(SYSTEM_POST_DATA_REQUIRED_VALUES_ONLY)
+        system_id = self.post_system(SYSTEM_POST_DATA_ALL_VALUES_NO_PARENT)
+        self.patch_system(system_id, {"name": SYSTEM_POST_DATA_REQUIRED_VALUES_ONLY["name"]})
+        self.check_patch_system_failed_with_message(
+            409, "A System with the same name already exists within the parent System"
+        )
```
joelvdavies, line 447: Also worth mentioning things like this I don't think are in the spirit of DSL, as really it would require additional functions with names such as

self.create_system_with_child()
self.move_system_to_previously_created_system()

But I thought that going that far is where the disadvantages of the method start to occur as maintenance becomes more awkward so I thought this could be a decent hybrid instead.

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
-    # Mock `create` to return the created System
-    test_helpers.mock_create(system_repository_mock, system)
+        self.test_helpers = test_helpers
+        self.mock_system_repository = system_repository_mock
+        self.system_service = system_service
 
-    created_system = system_service.create(SystemPostSchema(**system_info))
+        with patch("inventory_management_system_api.services.system.utils", wraps=utils) as wrapped_utils:
+            self.wrapped_utils = wrapped_utils
+            yield
```
VKTB, line 47: Not sure I know what the purpose of the wrapper here is but lets have a chat.

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
+        self.test_helpers.mock_get(self.mock_system_repository, self._expected_system)
 
-    # Mock `get` to return breadcrumbs
-    test_helpers.mock_get_breadcrumbs(system_repository_mock, breadcrumbs)
+    def call_get(self, system_id: str):
+        """Calls the SystemService `get` method"""
 
-    retrieved_breadcrumbs = system_service.get_breadcrumbs(system_id)
+        self._obtained_system_id = system_id
+        self._obtained_system = self.system_service.get(system_id)
```
VKTB, line 124: It's complaining about the type - `Expected type 'MagicMock', got 'SystemOut | None' instead `.

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
+
+        # Simply a return currently, so no need to use actual data
+        self._expected_breadcrumbs = MagicMock()
+        self.test_helpers.mock_get_breadcrumbs(self.mock_system_repository, self._expected_breadcrumbs)
+
+    def call_get_breadcrumbs(self, system_id: str):
+        """Calls the SystemService `get` method"""
+
+        self._obtained_system_id = system_id
+        self._obtained_breadcrumbs = self.system_service.get_breadcrumbs(system_id)
```
VKTB, line 163: It's complaining about the type - `Expected type 'MagicMock', got 'BreadcrumbsGetSchema' instead`

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
+
+        # Simply a return currently, so no need to use actual data
+        self._expected_systems = MagicMock()
+        self.test_helpers.mock_list(self.mock_system_repository, self._expected_systems)
+
+    def call_list(self, parent_id: Optional[str]):
+        """Calls the SystemService `list` method"""
+
+        self._parent_id_filter = parent_id
+        self._obtained_systems = self.system_service.list(parent_id)
```
VKTB, line 202: It's complaining about the type - `Expected type 'MagicMock', got 'list[SystemOut]' instead`

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
+        self._expected_system_in = SystemIn(
+            **merged_system_data,
+            code=utils.generate_code(merged_system_data["name"], "system"),
+        )
+
+    def call_update(self, system_id: str):
+        """Calls the SystemService `update` method with the appropriate data from a prior call to `mock_update`"""
+
+        self._updated_system_id = system_id
+        self._updated_system = self.system_service.update(system_id, self._system_patch)
```
VKTB, line 276: It's complaining about the type - `Expected type 'MagicMock', got 'SystemOut' instead `

test/unit/services/test_system.py
```diff
@@ -2,284 +2,374 @@
...
+
+    def call_update_expecting_error(self, system_id: str, error_type: type):
+        """Calls the SystemService `update` method with the appropriate data from a prior call to `mock_update`
+        while expecting an error to be raised"""
+
+        with pytest.raises(error_type) as exc:
```
VKTB, line 282: It's complaining about the type - `Expected type 'tuple[Any, ...] | Any' (matched generic type 'Type[E] | tuple[Type[E], ...]'), got 'type' instead`.

test/e2e/test_system.py
```diff
@@ -4,603 +4,579 @@
...
+        self.test_client = test_client
+        self._post_response = []
+
+    def post_system(self, system_post_data: dict) -> Optional[str]:
+        """Posts a System with the given data, returns the id of the created system if successful
+
+        :param system_post_data: Dictionary containing the system data that should be posted
+        :return: ID of the created system (or None if not successful)
+        """
+        self._post_response = self.test_client.post("/v1/systems", json=system_post_data)
```
VKTB, line 47: It's complaining about the type - `Expected type 'list[Response]', got 'Response' instead`.

test/e2e/test_system.py
```diff
@@ -4,603 +4,579 @@
...
+
+        return self._post_response.json()["id"] if self._post_response.status_code == 201 else None
+
+    def check_post_system_success(self, expected_system_get_data: dict):
+        """Checks that a prior call to 'post_system' gave a successful response with the expected data returned"""
+
+        assert self._post_response.status_code == 201
```
VKTB, line 54: There are quite a few of these - `Unresolved attribute reference 'status_code' for class 'list'`.

test/e2e/test_system.py
```diff
@@ -4,603 +4,579 @@
...
+        assert self._post_response.json() == expected_system_get_data
```
VKTB, line 55: There are quite a few of these - `Unresolved attribute reference 'json' for class 'list'`.

VKTB (commented):
This is very good work, thank you! It is so much more readable and I hope that it will be easier to maintain in future.

test/unit/repositories/test_system.py
```diff
@@ -24,779 +25,789 @@
...
+    def call_create(self):
+        """Calls the SystemRepo `create` method with the appropriate data from a prior call to `mock_create`"""
+
+        self._created_system = self.system_repository.create(self._system_in, session=self.mock_session)
+
+    def call_create_expecting_error(self, error_type: type):
+        """Calls the SystemRepo `create` method with the appropriate data from a prior call to `mock_create`
+        while expecting an error to be raised"""
+
+        with pytest.raises(error_type) as exc:
```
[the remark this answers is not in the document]
joelvdavies: Good spot, I think it either needs the generic type as pytest does or type[BaseException] seems to work. I will change my vscode settings to include type checking and go through them.

test/unit/repositories/test_system.py
```diff
@@ -24,779 +25,789 @@
...
+        self.mock_utils.compute_breadcrumbs.return_value = self._expected_breadcrumbs
 
+    def call_get_breadcrumbs(self, system_id: str):
+        """Calls the SystemRepo `get_breadcrumbs` method"""
 
-def test_update_duplicate_name_within_parent(test_helpers, database_mock, system_repository):
-    """
-    Test updating a System with a duplicate name within the parent System
+        self._obtained_system_id = system_id
+        self._obtained_breadcrumbs = self.system_repository.get_breadcrumbs(system_id, session=self.mock_session)
```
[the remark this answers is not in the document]
joelvdavies: I can't see this one any more, but I think it may have been fixed by resolving some other type errors. Could you recheck this once I have committed please.

test/unit/repositories/test_system.py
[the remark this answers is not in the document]
joelvdavies: I have resolved most but get ones for id= not existing because of the alias - I think this is alright though, not sure if yours shows those?

test/e2e/test_system.py
joelvdavies: I messed up the type while experimenting with merging some of the layers, should have just been Response not list[Response].

test/unit/services/test_system.py
joelvdavies: I should put a comment here.

test/unit/repositories/test_system.py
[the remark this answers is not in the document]
joelvdavies: Leaving id ones - seems to be only be found by vscode and not pycharm.

test/unit/repositories/test_system.py
```diff
@@ -24,779 +25,789 @@
...
+        """
+        inserted_system_id = CustomObjectId(str(ObjectId()))
+
+        # Pass through system_in first as need creation and modified times
+        self._system_in = SystemIn(**system_in_data)
+
+        self._expected_system_out = SystemOut(**self._system_in.model_dump(), id=inserted_system_id)
+
+        # When a parent_id is given, need to mock the find_one for it too
+        if system_in_data["parent_id"]:
```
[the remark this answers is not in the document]
joelvdavies: We discussed this and it is as expected.

test/e2e/test_system.py
```diff
@@ -4,603 +4,579 @@
...
-        **update_values,
-        **additional_expected_values,
-    }
+        assert self._get_response.status_code == 200
+        assert self._get_response.json() == {
+            "trail": [
+                [system["id"], system["name"]]
+                # When the expected trail length is < the number of systems posted, only use the last
+                for system in self._posted_systems_get_data[
+                    (len(self._posted_systems_get_data) - expected_trail_length) :
```
VKTB, line 219:
```suggestion
                    (len(self._posted_systems_get_data) - expected_trail_length):
```

joelvdavies: This is black's formatter.

VKTB: 🤦‍♂️ never seen this before :joy:

test/e2e/test_system.py
VKTB: Like you say, it won't be clear until we get that far.

test/unit/repositories/test_system.py
VKTB: As discussed, this can be done in a separate PR.

test/e2e/test_system.py
joelvdavies: Yeah me neither 😄

PR #343, opened 2024-07-31
Refactor CatalogueItemService unit tests #341

### LINKED ISSUE #341: Refactor catalogue items service unit tests
As part of https://github.com/ral-facilities/inventory-management-system-api/issues/90 following on from https://github.com/ral-facilities/inventory-management-system-api/issues/340.

### LINKED ISSUE #349: Error updating catalogue item's catalogue category to different number of properties
While working on #343 I found that when updating a catalogue item's `catalogue_category_id` to one that has no properties (from one that does) - the current check on equal properties doesn't work as the for loop never runs. This means it subsequently throws an error that isn't caught.

joelvdavies:
## Description
See #341.

### Notes
- Removed some unit tests that were already handled by `utils.process_properties`  e.g. `test_update_add_non_mandatory_property`, `test_update_remove_non_mandatory_property` etc.
- Removed `test_get_with_non_existent_id` - Logic handled in repo unit tests and e2e tests (matches already refactored service tests)
- Found and fixed #349 

I think we should be able to clean up the mock_create/update methods later by adding some additional functions in the tests/helper functions in conftest, although its not clear how it will impact items yet or the e2e tests yet.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #341, Closes #349

### DIFF
--- a/test/unit/services/test_catalogue_item.py
+++ b/test/unit/services/test_catalogue_item.py
@@ -669,7 +669,6 @@ class UpdateDSL(CatalogueItemServiceDSL):
             )
 
         # When properties are given need to add any property `id`s and ensure the expected data inserts them as well
-        property_post_schemas = []
         expected_properties_in = []
         if self._updating_properties:
 
@@ -756,8 +755,9 @@ class UpdateDSL(CatalogueItemServiceDSL):
             )
 
         # Ensure obtained new catalogue category if moving
+        expected_catalogue_category_get_calls = []
         if self._moving_catalogue_item and self._catalogue_item_patch.catalogue_category_id:
-            expected_catalogue_category_get_calls = [call(self._catalogue_item_patch.catalogue_category_id)]
+            expected_catalogue_category_get_calls.append(call(self._catalogue_item_patch.catalogue_category_id))
 
             # Expect additional call from existing catalogue category to compare properties if new properties aren't
             # given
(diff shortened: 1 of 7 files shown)


test/unit/services/test_catalogue_item.py
```diff
@@ -2,2109 +2,1200 @@
...
+        self.mock_catalogue_category_repository = catalogue_category_repository_mock
+        self.mock_manufacturer_repository = manufacturer_repository_mock
+        self.mock_unit_repository = unit_repository_mock
+        self.catalogue_item_service = catalogue_item_service
+
+        with patch("inventory_management_system_api.services.catalogue_item.utils", wraps=utils) as wrapped_utils:
+            self.wrapped_utils = wrapped_utils
+            yield
+
+    def construct_properties_in_and_post_with_ids(
```
MatteoGuarnaccia5, line 90: could all the stuff we do with properties be abstracted even more? In a separate file maybe so it can be used by catalogue categories, catalogue items and items

MatteoGuarnaccia5 (approved):
LGTM - just one comment

test/unit/services/test_catalogue_item.py
joelvdavies: Yeah good idea, Viktor and I tried it briefly and didn't like how it was turning out at the time, so decided to leave it until items as we thought it would be clearer how we can simplify these then.

VKTB (commented):
Just some very minor things, it looks good otherwise.

VKTB (approved):

PR #352, opened 2024-08-05
Refactor catalogue item e2e tests #342

### LINKED ISSUE #342: Refactor catalogue items e2e tests
As part of https://github.com/ral-facilities/inventory-management-system-api/issues/90 following on from https://github.com/ral-facilities/inventory-management-system-api/issues/341.

joelvdavies:
## Description
See #342.

### Notes
- Removed `test_create_catalogue_item_with_allowed_values` - effectively tested through other tests e.g. `test_create_with_all_properties_defined`.
- Adds a couple more create tests around properties
- For `test_partial_update_catalogue_category_id_no_properties` to work - needed to rename `_post_response` to reflect the actual entity posted to avoid confusion between the last category posted, and the last item
- Not sure about the `_GET_DATA` name anymore as they can be missing IDs but couldn't think of a nice alternative as _DATA is already taken for _POST so added a note
- We may wish, rather than trying to inherit multiple classes from manufacturers, systems and catalogue categories from the test files to use https://github.com/ral-facilities/inventory-management-system-api/issues/90#issuecomment-2064390649 to store these appropriate ids and handle the posting of data. Then again inheriting gives access through self. which I think would be desirable for consistency and otherwise would probably still need these functions, but posting via the client instead.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #342


test/unit/services/test_catalogue_category.py
```diff
@@ -484,7 +484,7 @@ def mock_update(
         :param catalogue_category_id: ID of the catalogue category that will be obtained.
         :param catalogue_category_update_data: Dictionary containing the basic patch data as would be required for a
                                                `CatalogueCategoryPatchSchema` but with any unit_id's replaced by the
```
MatteoGuarnaccia5, line 486:
```suggestion
                                               `CatalogueCategoryPatchSchema` but with any `unit_id`'s replaced by the
```

MatteoGuarnaccia5 (changes requested):
Looks good, just some typos in docstrings

test/e2e/test_catalogue_item.py
```diff
@@ -2,2198 +2,1539 @@
...
+    def post_manufacturer(self, manufacturer_post_data: dict) -> Optional[str]:
+        """
+        Posts a manufacturer with the given data, returns the ID of the created manufacturer if successful.
+
+        :param manufacturer_post_data: Dictionary containing the manufacturer data as would be required for a
+                                       `ManufacturerPostSchema`.
+        :return: ID of the created manufacturer (or `None` if not successful).
+        """
+        self.manufacturer_id = ManufacturerCreateDSL.post_manufacturer(self, manufacturer_post_data)
+        return self.manufacturer_id
```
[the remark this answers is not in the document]
joelvdavies: Yeah we could save the lines here, but then every time we do a post we would also have to assign the ID, I thought this way was slightly nicer given the ids are left to the function itself rather than being present in the test which I thought would be less obvious to the person writing the test. We may use a test client as described in #90 which again could do the same thing but that is to be decided later.

test/e2e/test_catalogue_item.py
```diff
@@ -2,2198 +2,1539 @@
...
+        self.manufacturer_id = "invalid-id"
+        self.post_catalogue_item(CATALOGUE_ITEM_DATA_REQUIRED_VALUES_ONLY)
+
+        self.check_post_catalogue_item_failed_with_detail(422, "The specified manufacturer does not exist")
+
+
+class GetDSL(CreateDSL):
+    """Base class for get tests."""
+
+    _get_response: Response
```
VKTB, line 584: Should this be `_get_response_catalogue_item` instead?

test/e2e/test_catalogue_item.py
```diff
@@ -2,2198 +2,1539 @@
...
+    """Base class for list tests."""
+
+    def get_catalogue_items(self, filters: dict) -> None:
+        """
+        Gets a list catalogue items with the given filters.
+
+        :param filters: Filters to use in the request.
+        """
+
+        self._get_response = self.test_client.get("/v1/catalogue-items", params=filters)
```
VKTB, line 659: Same here - `_get_response_catalogue_item`?

test/e2e/test_catalogue_item.py
```diff
@@ -2,2198 +2,1539 @@
...
+
+    def post_test_catalogue_items(self) -> list[dict]:
```
VKTB, line 661: Is `test` needed in the name of this method?

VKTB (commented):
Just some very minor things, it looks good otherwise.

VKTB:
> - We may wish, rather than trying to inherit multiple classes from manufacturers, systems and catalogue categories from the test files to use https://github.com/ral-facilities/inventory-management-system-api/issues/90#issuecomment-2064390649 to store these appropriate ids and handle the posting of data. Then again inheriting gives access through self. which I think would be desirable for consistency and otherwise would probably still need these functions, but posting via the client instead.

We should discuss this as I am not sure I fully understand you.

test/e2e/test_catalogue_item.py
joelvdavies: It matches the previous e2e tests as I thought `post_catalogue_items` is a bit confusing as it doesnt post something its given, although I admit the other names tend to be clearer e.g. `post_test_catalogue_category_with_child` so open to a better name. The only thing I can think of right now is just being more explicit e.g. `post_two_test_catalogue_items`.

VKTB: Oh I see what it was supposed to mean, I just thought it was a typo. Please ignore me and keep it as it is, sorry.

VKTB (approved):

joelvdavies:
> > * We may wish, rather than trying to inherit multiple classes from manufacturers, systems and catalogue categories from the test files to use [Clean up tests #90 (comment)](https://github.com/ral-facilities/inventory-management-system-api/issues/90#issuecomment-2064390649) to store these appropriate ids and handle the posting of data. Then again inheriting gives access through self. which I think would be desirable for consistency and otherwise would probably still need these functions, but posting via the client instead.
> 
> We should discuss this as I am not sure I fully understand you.

I think I was just suggesting an alternative method to the way I have essentially inherited and overridden some post methods e.g. `post_manufacturer` just to assign the manufacturer_id automatically when its called. I.e. we could also go via a test client that always does this which reduces the amount of inheritance down the line, but equally I wasn't convinced so I was just commenting on it so you would be aware.

PR #354, opened 2024-08-07
Refactor ItemRepo unit tests #345

### LINKED ISSUE #345: Refactor items repo unit tests
As part of #90 following on from #342.

joelvdavies:
## Description
See #345.

### Notes
- I noticed we test the `system_id` existence in the create, but not in the update (we do that in the service layer) - we may want to change this later for consistency
- Removed `test_list_with_invalid_system_id_filter` none of the other repo tests do invalid id checks
- Removed `test_with_catalogue_item_id_filter_no_matching_results` and `test_list_with_system_id_filter_no_matching_results` and similar - they are instead tested at once in `test_list_with_system_id_and_catalogue_category_id_with_no_results`

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #345

### DIFF
--- a/inventory_management_system_api/repositories/item.py
+++ b/inventory_management_system_api/repositories/item.py
@@ -51,20 +51,6 @@ class ItemRepo:
         item = self.get(str(result.inserted_id), session=session)
         return item
 
-    def delete(self, item_id: str, session: ClientSession = None) -> None:
-        """
-        Delete an item by its ID from a MongoDB database.
-
-        :param item_id: The ID of the item to delete.
-        :param session: PyMongo ClientSession to use for database operations
-        :raises MissingRecordError: If the item doesn't exist
-        """
-        item_id = CustomObjectId(item_id)
-        logger.info("Deleting item with ID: %s from the database", item_id)
-        result = self._items_collection.delete_one({"_id": item_id}, session=session)
-        if result.deleted_count == 0:
-            raise MissingRecordError(f"No item found with ID: {str(item_id)}")
-
     def get(self, item_id: str, session: ClientSession = None) -> Optional[ItemOut]:
         """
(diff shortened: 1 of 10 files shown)


VKTB:
> - I noticed we test the `system_id` existence in the create, but not in the update (we do that in the service layer) - we may want to change this later for consistency

Sorry but what are you suggesting we change for consistency? The reason we are asserting in the service unit tests for update is because the `system_id` existence check lives in the service layer. Are you suggesting that we keep the existence checks in one same layer?

VKTB (approved):
LGTM

joelvdavies:
> > * I noticed we test the `system_id` existence in the create, but not in the update (we do that in the service layer) - we may want to change this later for consistency
> 
> Sorry but what are you suggesting we change for consistency? The reason we are asserting in the service unit tests instead of the repo unit tests for update is because the `system_id` existence check lives in the service layer. Are you suggesting that we keep the existence checks in one same layer?

Yeah I think that's what I meant - again I think I was just voicing a thought as I went through them that we haven't been consistent.

VKTB:
> > > * I noticed we test the `system_id` existence in the create, but not in the update (we do that in the service layer) - we may want to change this later for consistency
> > 
> > 
> > Sorry but what are you suggesting we change for consistency? The reason we are asserting in the service unit tests instead of the repo unit tests for update is because the `system_id` existence check lives in the service layer. Are you suggesting that we keep the existence checks in one same layer?
> 
> Yeah I think that's what I meant - again I think I was just voicing a thought as I went through them that we haven't been consistent.

I think for some checks there are reasons why they are not in the same layer but for this one it seems like (at first glance) that it can be moved to one layer.

PR #393, opened 2024-10-09
Add Expected Lifetime to Catalogue Items

### LINKED ISSUE #338: Add expected lifetime field to catalogue items
Catalogue items should have an optional `expected_lifetime` field. A migration script should also be written to back populate existing data.

asuresh-code:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes #338


joelvdavies (changes requested):
The migration script doesn't seem to work on the mock data, I get several
image errors.

inventory_management_system_api/migrations/scripts/expected_lifetime_migration.py
```diff
@@ -0,0 +1,211 @@
...
+                    "is_obsolete": old_item.is_obsolete,
+                    "obsolete_reason": old_item.obsolete_reason,
+                    "obsolete_replacement_catalogue_item_id": old_item.obsolete_replacement_catalogue_item_id,
+                    "notes": old_item.notes,
+                    "properties": old_item.properties,
+                    "created_time": catalogue_item["created_time"],
+                    "modified_time": catalogue_item["modified_time"],
+                }
+
+                new_item = NewCatalogueItemIn(**new_item_data)
```
joelvdavies, line 188: You can use `.model_dump()` on the old model as well e.g.
```
new_item = NewCatalogueItemIn(**old_item.model_dump(), expected_lifetime=None)
```

Then you can use something like
```
update_data = {**new_item.model_dump(), "modified_time":  old_item.modified_time}

File: inventory_management_system_api/migrations/scripts/expected_lifetime_migration.py [hunk 2]
```diff
@@ -0,0 +1,188 @@
...
+                self._catalogue_items_collection.replace_one(
+                    {"_id": catalogue_item["_id"]},
+                    update_data,
+                    session=session,
+                )
+
+            except ValidationError as ve:
+                logger.error("Validation failed for item with id %s: %s", catalogue_item["_id"], ve)
+
+                continue
```
joelvdavies, line 176: This shouldn't either shouldn't be caught here or you need to raise the exception again e.g. `raise ve` - the way the migration script works is that it runs `forward` in a transaction meaning that if an error occurs and isn't caught, then the whole migration reverts. Whereas by catching it will carry on meaning the database could end up with some valid documents migrated, and some invalid ones not.

inventory_management_system_api/migrations/scripts/expected_lifetime_migration.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __init__(self, database: Database):
+
+        self._catalogue_items_collection: Collection = database.catalogue_items
+
+    def forward(self, session: ClientSession):
+        """This function should actually perform the migration
+
+        All database functions should be given the session in order to ensure all updates are done within a transaction
+        """
```
joelvdavies, line 152: This comment should probably be updated (same for backward).

joelvdavies (commented):
Just two comments, script worked well for me 🥳

inventory_management_system_api/migrations/scripts/expected_lifetime_migration.py
asuresh-code: Ah, okay I think I understand. The error should be uncaught, so that the transaction gets terminated, because we don't some documents with the field and others missing it. So I can either raise another error, or just remove the `except`.

joelvdavies: Yes exactly

joelvdavies (approved):
LGTM

VKTB (approved):
LGTM 🎉 Thank you for your work!

joelvdavies (approved):
This can be safely merged prior to https://github.com/ral-facilities/inventory-management-system/issues/1020 if you wish. The extra field will just be ignored.

PR #421, opened 2024-11-19
Add endpoint to set spares definition #413

### LINKED ISSUE #413: Add endpoint to set the definition of a spare
The definition of a spare is effectively a list of usage statuses that a spare would have. We need an endpoint that allows the front end to assign this definition. Endpoint should:
- [x] Be a PUT at `/settings/spares_definition`
- [x] Take a list of usage statuses that constitute a spare e.g.
   ```
  {
    "usage_statuses": [
      { "id": "6723a734ae807034e53cdd13" },
      { "id": "6723a734ae807034e53cdd14" }
    ]
  }
   ```
- [x] Store these usage statuses for later retrieval
   - Use a `settings` database collection with an `_id: "spares_definition"`
   - Either store the values like properties or we can insert them later (e.g. aggregation)
- [x] This task should also ensure ensure a usage status cannot be deleted while it is in the definition

joelvdavies:
## Description

See #413. Went with the aggregate query approach as an experiment, we could equally just loop though and insert the values. There is no need to insert all usage status fields as I have done at the moment.

### Notes

- Had a conflict with some of the pytest `setup` fixtures when inheriting so renamed all in e2e tests to be consistent
- Moved some id_dicts to their base class e.g. instead of having `usage_status_value_id_dict` in catalogue category's create dsl, it is now populated in the usage status create dsl to avoid repeating in cases like this

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #413, relates to #90


asuresh-code (changes requested):
Looks Good overall, and seems to work well. Have some minor changes/comments on the implementation, testing is good except for a small typo.

asuresh-code:
In the Settings Schema, it says L54 is not covered by tests, although I can see there is a `duplicate_usage_status` test that checks for the error, so I'm not sure if I'm missing something or the tool has made a mistake.
image

inventory_management_system_api/schemas/setting.py
```diff
@@ -0,0 +1,65 @@
...
+
+    usage_statuses: conlist(SparesDefinitionPutUsageStatusSchema, min_length=1) = Field(
+        description="Usage statuses that classify items as a spare."
+    )
+
+    @field_validator("usage_statuses")
+    @classmethod
+    def validate_usage_statuses(
+        cls, usage_statuses: list[SparesDefinitionPutUsageStatusSchema]
+    ) -> list[SparesDefinitionPutUsageStatusSchema]:
```
asuresh-code, line 39: The `usage_statuses` param is type casted to `SparesDefinitionPutUsageStatusSchema`, but when it's reused in the `SparesDefinitionSchema`, its type for `usage_statuses` is `UsageStatusSchema`. There doesn't seem to be any errors caused by this, since they both have the same id field, but the param's type could be changed to a union of the 2 schemas.

test/unit/repositories/test_usage_status.py
```diff
@@ -385,6 +395,16 @@ def mock_is_usage_status_in_item(self, item_data: Optional[dict] = None) -> None
         self._mock_item_data = item_data
         RepositoryTestHelpers.mock_find_one(self.items_collection, item_data)
 
+    def mock_is_usage_status_in_setting(self, setting_data: Optional[dict] = None) -> None:
+        """
+        Mocks database methods appropriately for when the `_is_usage_status_in_setting` repo method will be
+        called.
+
+        :param setting_data: Dictionary containing an settings's data (or `None`).
```
asuresh-code, line 403:
```suggestion
        :param setting_data: Dictionary containing a settings's data (or `None`).
```

inventory_management_system_api/schemas/setting.py
joelvdavies: True, good spot! I think looking over the others, this is the only case where it occurs, I think its best to avoid having the validator on the get though as we only really wanted it on the post in the first place. In this case the simplest solution is to just inherit from `BaseModel` instead of `SparesDefinitionPutSchema` as its only overwriting the one field anyway.

inventory_management_system_api/repositories/setting.py
```diff
@@ -0,0 +1,107 @@
...
+from pymongo.collection import Collection
+
+from inventory_management_system_api.core.database import DatabaseDep
+from inventory_management_system_api.models.setting import SettingInBase, SettingOutBase, SparesDefinitionOut
+
+logger = logging.getLogger()
+
+# Template types for models inheriting from SettingIn/OutBase so this repo can be used generically for multiple settings
+SettingInBaseT = TypeVar("SettingInBaseT", bound=SettingInBase)
+SettingOutBaseT = TypeVar("SettingOutBaseT", bound=SettingOutBase)
```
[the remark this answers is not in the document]
joelvdavies: I think having the union here would actually mean it can be any subtype of SettingOutBase or SettingInBase (https://stackoverflow.com/questions/59933946/difference-between-typevart-a-b-and-typevart-bound-uniona-b). I don't think there is a way to have them coupled like that unfortunately.

image

joelvdavies:
> `duplicate_usage_status`

Yes codecov is correct as it excludes e2e tests in its coverage checking. We have generally ignored unit testing validation in the schemas in unit tests due to them usually being created inside the router level when a request is made, so without adding a test file for schemas it makes sense to use the e2e tests to validate them instead but as a consequence they are left untested in unit tests and so show up on the coverage e.g. https://app.codecov.io/gh/ral-facilities/inventory-management-system-api/pull/421/blob/inventory_management_system_api/schemas/catalogue_category.py.

asuresh-code (changes requested):
Looks good overall, just some more final minor changes.

inventory_management_system_api/repositories/setting.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        self._database = database
+        self._settings_collection: Collection = self._database.settings
+
+    def upsert(
+        self, setting: SettingInBaseT, out_model_type: Type[SettingOutBaseT], session: ClientSession = None
+    ) -> SettingOutBaseT:
+        """
+        Assign a setting a MongoDB database. Will either update or insert the setting depending on whether it
+        already exists.
```
asuresh-code, line 72:
```suggestion
        Update or insert the setting depending on whether it already exists.
```
Not sure what it means by `Assign a setting a MongoDB database`. If it's a mistake, I think this is a good alternative comment

inventory_management_system_api/repositories/setting.py
```diff
@@ -0,0 +1,107 @@
...
+        Initialize the `SettingRepo` with a MongoDB database instance.
+
+        :param database: The database to use.
+        """
+        self._database = database
+        self._settings_collection: Collection = self._database.settings
+
+    def upsert(
+        self, setting: SettingInBaseT, out_model_type: Type[SettingOutBaseT], session: ClientSession = None
+    ) -> SettingOutBaseT:
```
asuresh-code, line 69:
```suggestion
    def upsert(
        self, setting: SettingInBaseT, out_model_type: Type[SettingOutBaseT], session: ClientSession = None
    ) -> Optional[SettingOutBaseT]:
```
The `get` used in the return of this function has an optional return type. We do expect a `setting` since we just upserted one previously, but technically by the get function's defintion, the return value could be `None`, so I'm not sure whether an Optional is necessary or not for the return type.

joelvdavies: Yeah I opted for not having it because the get in this case comes after the upsert so it should always exist in this context. I believe we do the same on the other updates for the same reason.

asuresh-code (approved):
LGTM!

inventory_management_system_api/repositories/usage_status.py
```diff
@@ -115,11 +123,25 @@ def _is_duplicate_usage_status(
         return usage_status is not None
 
     def _is_usage_status_in_item(self, usage_status_id: str, session: ClientSession = None) -> bool:
```
VKTB, line 125:
```suggestion
    def _is_usage_status_in_item(self, usage_status_id: CustomObjectId, session: ClientSession = None) -> bool:
```

inventory_management_system_api/services/setting.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Initialise the `SettingService` with a `SettingRepo` repository.
+
+        :param setting_repository: `SettingRepo` repository to use.
+        :param usage_status_repository: `UsageStatusRepo` repository to use
+        """
+        self._setting_repository = setting_repository
+        self._usage_status_repository = usage_status_repository
+
+    def set_spares_definition(self, spares_definition: SparesDefinitionPutSchema) -> SparesDefinitionOut:
```
VKTB, line 38: `set_` or `update_`?

inventory_management_system_api/services/setting.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Sets the spares definition to a new value.
```
VKTB, line 40: `Sets` or `Updates`?

VKTB (changes requested):
Works well, just some minor things.

inventory_management_system_api/routers/v1/setting.py
```diff
@@ -0,0 +1,39 @@
...
+
+SettingServiceDep = Annotated[SettingService, Depends(SettingService)]
+
+
+@router.put(
+    path="/spares_definition",
+    summary="Set the definition of a spare",
+    response_description="Spares definition updated successfully",
+)
+def set_spares_definition(
```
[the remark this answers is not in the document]
joelvdavies: I used set as it felt like it covered both update and create, and also thinking of it in the context of the setting existing but having a `None` value when it hasn't been inserted yet.

inventory_management_system_api/schemas/setting.py
```diff
@@ -0,0 +1,64 @@
...
+
+    id: str = Field(description="The ID of the usage status.")
+
+
+class SparesDefinitionPutSchema(BaseModel):
+    """
+    Schema model for a spares definition update request.
+    """
+
+    usage_statuses: conlist(SparesDefinitionPutUsageStatusSchema, min_length=1) = Field(
```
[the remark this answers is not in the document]
joelvdavies: Ah nice, I didn't know that, it gets around https://github.com/microsoft/pylance-release/issues/5457 as well, at least where we use fields anyway.

inventory_management_system_api/schemas/setting.py
```diff
@@ -0,0 +1,64 @@
...
+
+    id: str = Field(description="The ID of the usage status.")
+
+
+class SparesDefinitionUsageStatusSchema(SparesDefinitionPutUsageStatusSchema):
+    """
+    Schema model for a usage status within a spares definition.
+    """
+
+    id: str = Field(description="The ID of the usage status.")
```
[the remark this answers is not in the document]
joelvdavies: 🤦‍♂️ Yep seems I wrote it twice.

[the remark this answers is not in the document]
joelvdavies: 🤦‍♂️ Its a remnant from when I originally had it returning just the id and eventually the value, before I decided to just take the whole usage status.

inventory_management_system_api/routers/v1/setting.py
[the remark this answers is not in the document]
joelvdavies: Although then again even in that context it works, its always updating the setting whether it is already in the database or not, and matches the mongodb update. I will change it.

VKTB (approved):

PR #423, opened 2024-11-22
Add endpoint to get spares definition #414

### LINKED ISSUE #414: Add endpoint to get the definition of a spare
Requires #413. Endpoint should:
- [x] Be a GET at `/settings/spares_definition`
- [x] Return the list of usage statuses that are classed as spares e.g.
    ```
    {
     "usage_statuses": [
       { "id": "6723a734ae807034e53cdd13", value: "In Use" },
       { "id": "6723a734ae807034e53cdd14", value: "Using" }
     ]
    }
    ```

joelvdavies:
## Description
See #414. Follows on from #413.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #414


inventory_management_system_api/routers/v1/setting.py
```diff
@@ -37,3 +37,16 @@ def update_spares_definition(
...
+    path="/spares_definition", summary="Get the definition of a spare", response_description="Spares definition"
+)
+def get_spares_definition(setting_service: SettingServiceDep) -> SparesDefinitionSchema:
+    # pylint: disable=missing-function-docstring
+    logger.info("Getting spares definition")
+
+    spares_definition = setting_service.get_spares_definition()
+    if not spares_definition:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Spares definition not found.")
+    return spares_definition
```
VKTB, line 52: I see this warning in Pycharm `Expected type 'SparesDefinitionSchema', got 'SparesDefinitionOut' instead`.

VKTB (dismissed):

inventory_management_system_api/routers/v1/setting.py
joelvdavies: Ah yeah good spot 😄, fastapi handles the conversion the the return type automatically so it didn't break but we always convert them ourselves.

VKTB (approved):

PR #426, opened 2024-11-25
Add `number_of_spares` field to catalogue items #416

### LINKED ISSUE #416: Add `number_of_spares` field to catalogue items
This should
- [x] Default to `None`
- [x] Only be in the database model and get schemas - it should not be possible to set it on creation/edits (it will be calculated later)
- [x] Include a migration to add the field to all existing catalogue items (Again defaulting to `None`)

This is a simpler approach than the alternative of having the field non-optional at which point we would need the migration to also know the spares definition at the time of execution and would need all item creation requests to be blocked by the spares definition if it is not assigned.

joelvdavies:
## Description
See #416.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes #416


asuresh-code (changes requested):
Looks good overall, have some minor comments to make.

inventory_management_system_api/schemas/catalogue_item.py
```diff
@@ -95,3 +95,8 @@ class CatalogueItemSchema(CreatedModifiedSchemaMixin, CatalogueItemPostSchema):
         description="The properties specific to this catalogue item as defined "
         "in the corresponding catalogue category"
     )
+
+    # Computed
+    number_of_spares: Optional[int] = Field(
+        default=None, description="Number of spares currently available within this catalogue item if known"
```
asuresh-code, line 101:
```suggestion
        default=None, description="The number of spares currently available within this catalogue item"
```
All of the other descriptions start with `The` and `if known` doesn't seem necessary since other optional fields don't include this text.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -209,7 +210,7 @@ def update(self, catalogue_item_id: str, catalogue_item: CatalogueItemPatchSchem
 
         return self._catalogue_item_repository.update(
             catalogue_item_id,
-            CatalogueItemIn(**{**stored_catalogue_item.model_dump(), **update_data}),
+            CatalogueItemIn(**{**stored_catalogue_item.model_dump(), **update_data, "number_of_spares": None}),
```
asuresh-code, line 213:
```suggestion
            CatalogueItemIn(**{"number_of_spares": None, **stored_catalogue_item.model_dump(), **update_data}),
```
This prevents any existing `number_of_spares` data from being reset to `None`. However, I don't know if you want to invalidate any spares calculations after you update Catalogue item data. In that case this change isn't necessary.

asuresh-code:
More of a general comment about the migration scripts, but we now have 2 separate migration scripts for catalogue items, for the `expected_lifetime` and `number_of_spares` fields respectively. If we do migrations with the `expected_lifetime` script on the database currently, wouldn't they fail, since there will be a new field `number_of_spares` that would cause the `OldCatalogueItem` and `NewCatalogueItem` models to fail?

inventory_management_system_api/schemas/catalogue_item.py
joelvdavies: True, the others I have done have not as I thought it was unnecessary to have the `The`, but I will for the sake of keeping it the same in the file, the rest can be unified later. I think the `if known` is worth emphasising in this case? They others are directly edited by the user, so I expect if there is an optional `description` then they would understand that when its None it just doesn't have one rather than it not being known, whereas the spares it is specifically because it isn't known not just that it doesn't have one.

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Yeah not sure why I even put it there, it should be unchanged as its the items that cause the recalculation 😄 good spot.

joelvdavies:
> More of a general comment about the migration scripts, but we now have 2 separate migration scripts for catalogue items, for the `expected_lifetime` and `number_of_spares` fields respectively. If we do migrations with the `expected_lifetime` script on the database currently, wouldn't they fail, since there will be a new field `number_of_spares` that would cause the `OldCatalogueItem` and `NewCatalogueItem` models to fail?

Yes that is correct, the migrations have to be done in the right order at the right time. I had the same thought and created another PR to implement something like alembic https://github.com/ral-facilities/inventory-management-system-api/pull/430 to address these.

inventory_management_system_api/schemas/catalogue_item.py
asuresh-code: Yeah I think it makes sense to leave `if known` in for this case. I thought there were some fields similar to `number_of_spares` (Optional, not known but exists), but I just realised I was looking at the Patch Schema where everything is made Optional

asuresh-code (changes requested):
Just some final comments in regards to the mock_data

test/mock_data.py
```diff
@@ -494,6 +494,7 @@
     **CATALOGUE_ITEM_DATA_REQUIRED_VALUES_ONLY,
     "catalogue_category_id": str(ObjectId()),
     "manufacturer_id": str(ObjectId()),
+    "number_of_spares": None,
 }
```
asuresh-code, line 498:
```suggestion
    **CATALOGUE_ITEM_DATA_REQUIRED_VALUES_ONLY,
    "catalogue_category_id": str(ObjectId()),
    "manufacturer_id": str(ObjectId()),
}
```
It's optional, so it shouldn't be included in the `REQUIRED_VALUES_ONLY`?

asuresh-code:
In the `mock_data.py`, `number_of_spares` is included in the `INDATA` and `GETDATA` catalogue item entities, but not the `DATA` catalogue item entities. I understand why it's included in the `GETDATA` ones, but shouldn't they be omitted from the `INDATA` examples, since they are not inputted during Catalogue Item Patch/Post requests? Similarly, I assumed that `DATA` entities represent the attributes catalogue items possess when stored in the database, in which case shouldn't they include `number_of_spares`?

test/mock_data.py
joelvdavies: The reason it is like this is because the `CatalogueItemBase` doesn't have it set to `None` automatically. The idea being that with the actual calculation, I never want it to default to anything - it should only be None if the calculation itself returns it, but forgetting to set it will cause an error and prompt a developer to use the calculation to populate it instead.  The others are set to None because if you forget to set it, its fine.

joelvdavies:
> `number_of_spares`

Yes you are write about the in data, they 

> In the `mock_data.py`, `number_of_spares` is included in the `INDATA` and `GETDATA` catalogue item entities, but not the `DATA` catalogue item entities. I understand why it's included in the `GETDATA` ones, but shouldn't they be omitted from the `INDATA` examples, since they are not inputted during Catalogue Item Patch/Post requests? Similarly, I assumed that `DATA` entities represent the attributes catalogue items possess when stored in the database, in which case shouldn't they include `number_of_spares`?

I have put some comments at the top of the mock_data.py to try and explain these models. All the `_IN_DATA`'s should match the `In` models, so they reflect what should go into the repository methods and into the database, which in this case is different to the patch/post schemas because they are not user assignable and are only calculated internally (similar to the comment I made above). The `_DATA` only dictionaries are for post schemas, and I don't think I do add the `number_of_spares` to them? You are right that those shouldn't because a user can't assign them. The `_GET_DATA` on the other hand does because a user getting a catalogue item needs to see the number of spares in the returned schema (its view only).

asuresh-code (approved):

VKTB (approved):

PR #430, opened 2024-11-25
Update migration script to handle multiple migrations automatically #425

### LINKED ISSUE #425: Improve migration script for multiple migrations
The current migration script was made to do one migration at a time, manually running each in succession. I think ideally when there are multiple it should be possible to run more than one in the same transaction e.g. for a v2 database upgrade.

Some potential ideas:
- Dates in migrations - to indicate when a migration was added allowing an admin to see which migrations are likely needed for an update and in which order they need to be executed
- Allow multiple migrations in the same transaction - so if the order is wrong/interaction between them fails the database is left unchanged
- Store some sort of indication of the last migration executed stored in the database like alembic would to keep track of what migrations actually need to be run (can use date for order instead of hash)

joelvdavies:
## Description

See #425. In its current form this PR replicates similar behaviour to alembic. New migrations can be added using
```bash
ims-migrate generate test_migration "Does nothing"
```
Then the last migration applied is stored inside the database, and looked up to determine which migrations need to be applied in order to reach a desired one e.g.
```bash
ims-migrate forward latest
```
Applies all unapplied forward migrations in a single transaction. (This asks first although it would be simple to add a -y and even do what others do with alembic and put this in the docker file on start up - but its only a single command now anyway.)

This should hopefully make it easier to to apply the required migrations in the right order without specifying a list of migrations to perform when its time to update production.

### Notes
- The changes applied outside of a transaction e.g. dropping collections still occurs outside of the transaction but will be applied after all of the migrations have inside so these operations will technically be out of order

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test creating a new migration and migrating forward and backwards locally

## Agile board tracking

Closes #425


asuresh-code (changes requested):
Looks good overall and works well. Just have some comments to make on core.py and some minor ones for the test.

test/unit/migrations/test_core.py
```diff
@@ -0,0 +1,424 @@
...
+        """
+
+        self._mock_load_migration.assert_has_calls(
+            [call(migration_name) for migration_name in expected_migration_names]
+        )
+        assert self._obtained_migrations_forward == {
+            migration_name: self._mock_load_migration.return_value for migration_name in expected_migration_names
+        }
+
+    def check_load_migrations_forward_to_success_failed_with_exception(self, message: str) -> None:
```
asuresh-code, line 105:
```suggestion
    def check_load_migrations_forward_to_failed_with_exception(self, message: str) -> None:
```

test/unit/migrations/test_core.py
```diff
@@ -0,0 +1,424 @@
...
+
+        self._mock_load_migration.assert_has_calls(
+            [call(migration_name) for migration_name in expected_migration_names]
+        )
+        assert self._obtained_migrations_backward == (
+            {migration_name: self._mock_load_migration.return_value for migration_name in expected_migration_names},
+            expected_final_previous_migration_name,
+        )
+
+    def check_load_migrations_backward_to_success_failed_with_exception(self, message: str) -> None:
```
asuresh-code, line 243:
```suggestion
    def check_load_migrations_backward_to_failed_with_exception(self, message: str) -> None:
```

inventory_management_system_api/migrations/core.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    Returns a list of migrations backward that need to be applied to get from the last migration applied to the database
+    to the given one inclusive.
+
+    :param name: Name of the last migration backward to apply.
+    :returns: Tuple containing:
+              - List of dicts containing the names and instances of the migrations that need to be applied in the order
+                they should be applied.
+              - Either the name of the last migration before the one given or `None` if there aren't any.
+    """
```
asuresh-code, line 136: The backwards migration reverses the specified migration (since it is inclusive). Arguably it could be better set to exclusive instead, but it depends on what use cases you imagine it would be used for? Or if you just wanted to maintain consistency with the forwards migration, that would be reasonable as well.

joelvdavies: Yeah I had the same thought, the reason I didn't in the end is to keep consistency with alembic as it downgrades inclusive of the revision as this PR is effectively recreating it for mongodb. Arguably I could even replace the command names so that the current `forward` -> `upgrade`, `backward` -> `downgrade`, `set` -> `stamp`, `latest` -> `head`. But I left it with what we were already using for the forward and backward, and head doesn't make as much sense here as its not quite fully the same.

asuresh-code (changes requested):
Just some more minor docstring comments.

README.md
```diff
@@ -329,30 +329,64 @@ a microservice that provides user authentication against an LDAP server and retu
...
+```bash
+ims-migrate create <migration_name> <migration_description>
+```
+
+to create a new one inside the `inventory_management_system/migrations/scripts` directory. Then add the code necessary
+to perform the migration. See `_example_migration.py` for an example on how to implement one.
+
+#### Performing forward migrations
+
+Before performing a you can first check the current status of the database and any outstanding migrations using
```
asuresh-code, line 345:
```suggestion
Before performing a migration you can first check the current status of the database and any outstanding migrations using
```

asuresh-code:
Also, the `base.py` file is missing full stops in most of the docstring comments.

asuresh-code (approved):

inventory_management_system_api/migrations/core.py
```diff
@@ -0,0 +1,223 @@
...
+    to the given one inclusive.
+
+    :param name: Name of the last migration forward to apply. 'latest' will just use the latest one.
+    :returns: List of dicts containing the names and instances of the migrations that need to be applied in the order
+              they should be applied.
+    """
+
+    available_migrations = find_available_migrations()
+
+    start_index = 0
```
VKTB, line 99:
```suggestion
```
Line 159 will not run if line 145 does not run because it will exit.

README.md
```diff
@@ -329,30 +329,65 @@ a microservice that provides user authentication against an LDAP server and retu
...
 ```bash
-ims-migrate list
+ims-migrate status
 ```
 
-or
+or in Docker
 
 ```bash
 docker exec -it inventory_management_system_api_container ims-migrate list
```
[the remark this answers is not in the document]
joelvdavies: Should have been status.

VKTB (approved):

PR #460, opened 2025-02-13
Delete related attachment & images when an entity is deleted

### LINKED ISSUE #399: Delete related attachment & images when an entity is deleted
When a catalogue item, item or system is deleted. Any attachments/images should also be deleted. This relies on
- https://github.com/ral-facilities/object-storage-api/issues/76
- https://github.com/ral-facilities/object-storage-api/issues/104

I think the simplest method for authentication is to ensure that on the delete request ims-api does the following:
- Send a request to delete all related entity data using the same jwt token and wait for the response and then
   - If successful: Delete the entity
   - If authentication fails producing a 403: Replicate the same error so the front end does the token refresh
   - If any other error occurs or a response times out: Raise an appropriate error - Will likely need additional logic to prevent existing tests from failing due to not being able to contact the object-storage-api.

Further considerations
- May require integration tests

VKTB:
## Description
This PR introduces two new config values for the API to connect to and interact with an Object Storage API instance. I opted for a boolean ENV VAR (`OBJECT_STORAGE__ENABLED`) because this API can be run on its own without an Object Storage API instance. This was the simplest way to ensure the tests did not break as an Object Storage API instance will otherwise be required to make them pass. **These new config values will need to be added to the Docker IMS deployment configurations.**

I implemented an `ObjectStorageAPIClient` class whose methods are used in the services layer by the catalogue item, item, and system modules to delete any attachments and images associated to the entity that is being deleted using the same JWT access token supplied in the request. The entities in the IMS API database get deleted after the attachments and images are successfully deleted from the object storage. Unfortunately, there is no way to restore these if the delete database operation is not successful. Because the JWT access tokens have an expiry time, there could be cases where they are valid when received by the IMS API but expire when they are received by the Object Storage API. In cases like these, the IMS API will just return `403 Invalid token or expired token` so the consumer will just need to retry their request with a valid access token. The IMS API will return `500 Something went wrong` for any other errors that occur while communicating with the Object Storage API.

## Testing instructions
You will need an Object Storage API instance running with some attachments and images in S3.

- [ ] Review code
- [ ] Check Actions build
- [ ] Attachments and/or images in Object Storage API (and S3!) get deleted when an entity (catalogue item, item, and system) gets deleted.
- [ ]`204` returned when there are no attachments and/or images associated with the entity that is being deleted.

## Agile board tracking
closes #399


inventory_management_system_api/auth/jwt_bearer.py
```diff
@@ -43,6 +43,8 @@ async def __call__(self, request: Request) -> str:
         if not self._is_jwt_access_token_valid(credentials.credentials):
             raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Invalid token or expired token")
 
+        request.state.token = credentials.credentials
```
VKTB, line 46: Not sure if this is a nice thing to do as there are not many examples when I searched for it on GitHub. The alternative would be to do `request.headers.get("Authorization").split("Bearer ")[1]` in the endpoints.

inventory_management_system_api/core/object_storage_api_client.py
```diff
@@ -0,0 +1,61 @@
...
+
+        :param endpoint: The Object Storage API endpoint to send the request to.
+        :param access_token: The JWT access token for auth with the Object Storage API.
+        :param entity_id: The ID of the entity whose objects should be deleted.
+        :raises ObjectStorageAPIAuthenticationError: If auth with the Object Storage API fails.
+        :raises ObjectStorageAPIServerError: If any other error is encountered while communicating with the Object
+            Storage API.
+        """
+        url = f"{config.object_storage.api_url}{endpoint}"
+        headers = {"Authorization": f"Bearer {access_token}"}
```
joelvdavies, line 49: Should this logic be optional for when auth is disabled?

joelvdavies (commented):
Works well, I did realise however that doing the delete of the images/attachments first means that when a system for example cant be deleted due to having child elements, it deletes the attachments and images while the front end says it cant be deleted at all. Perhaps we should do this check before the delete request is sent to the object storage?

inventory_management_system_api/core/object_storage_api_client.py
VKTB: The thing is that theoretically auth can be disabled in the IMS API but enabled in the Object Storage API so requests will error. If we are okay with this then I can change it to be optional.

VKTB:
> Works well, I did realise however that doing the delete of the images/attachments first means that when a system for example cant be deleted due to having child elements, it deletes the attachments and images while the front end says it cant be deleted at all. Perhaps we should do this check before the delete request is sent to the object storage?

Hmm very good point, I didn't think about it.. We will either have to check for child elements or use transactions. If we go with the former then we have to move the child elements check to the service layer. Idk what will happen if the child element check happens, someone else adds a child system while the images and attachments are being deleted, and the operation to delete the system happens after this.

inventory_management_system_api/core/object_storage_api_client.py
joelvdavies: I would personally think if you are using it with authentication, they should both be with authentication. If we only needed to do something for a particular kind of test I would think its fine to just manually comment out the line kind of like we do testing extremely short refresh tokens/checking front end auth by returning specific responses.

joelvdavies:
> > Works well, I did realise however that doing the delete of the images/attachments first means that when a system for example cant be deleted due to having child elements, it deletes the attachments and images while the front end says it cant be deleted at all. Perhaps we should do this check before the delete request is sent to the object storage?
> 
> Hmm very good point, I didn't think about it.. We will either have to check for child elements or use transactions. If we go with the former then we have to move the child elements check to the service layer. Idk what will happen if the child element check happens (and there are no child elements at that point in time), someone else adds a child system while the images and attachments are being deleted, and the operation to delete the system happens after this.

In this case a transaction wouldn't work as its cross service. Good point about the concurrency, hadn't realised that but all these checks could get caught up like that as we don't do any sort of write locking between the check and delete for the element itself. Worst case we could technically have some orphaned children without doing something like that although admittedly I would consider it extremely unlikely in practice. Due to the cross service nature, I think its still best to minimise orphaned attachments & images myself over the reverse, so would keep the os-api request before the entity deletion regardless of if it uses a transaction or not. I think this means we have no choice but to move the child check into the service layer regardless of using a transaction or not.

asuresh-code (commented):
Currently, entities are deleted first from the Object Storage Database, and then the IMS Database. I understand that doing it in this order prevents orphaning of images/attachments, however, would it be better to delete from the IMS database first?

The main benefit I see, is that if someone deletes an entity, and it fails in the second database deletion, It will still appear deleted for them since it's no longer in the IMS database.

So there would be an improvement in the user experience, at the cost of orphaning files (which I imagine you could address by periodically checking and deleting any files with an entity ID that no longer exists in the IMS database).

inventory_management_system_api/core/object_storage_api_client.py
VKTB: I addressed this, cheers. What about timeout for the request itself? Pylint complains about it by default hence why I put `# pylint: disable=missing-timeout`.

VKTB:
After some discussion, we decided the logic to be in the following order:
1. Check for child elements in the service layer
2. Delete any attachments and images associated with the entity from Object Storage API
3. Delete the entity itself from IMS API

inventory_management_system_api/core/object_storage_api_client.py
joelvdavies: 10 seconds would be more than enough I would think, pylint just likes it to have something so doesn't hang forever, so as long as its longer than we would ever expect it to take its fine. If we expect it will take longer if there's more to delete then maybe even 60 to be safe.

test/unit/core/test_object_storage_api_client.py
```diff
@@ -0,0 +1,77 @@
...
+    @pytest.mark.parametrize(
+        "method, endpoint",
+        [
+            (ObjectStorageAPIClient.delete_attachments, "/attachments"),
+            (ObjectStorageAPIClient.delete_images, "/images"),
+        ],
+    )
+    @patch("inventory_management_system_api.core.object_storage_api_client.requests.delete")
+    def test_delete_success(self, mock_delete, method, endpoint):
+        """Test `delete_attachments` and `delete_images methods when Object Storage API responds with 204 No Content."""
```
rowan04, line 29:
```suggestion
        """Test `delete_attachments` and `delete_images` methods when Object Storage API responds with 204 No Content."""
```

test/unit/repositories/test_system.py
```diff
@@ -915,3 +858,88 @@ def test_delete_invalid_id(self):
...
+
+    def test_has_child_elements_with_child_system(self):
+        """Test `has_child_elements` when there is a child system but no child items."""
+
+        self.mock_has_child_elements(child_system_data=SYSTEM_IN_DATA_NO_PARENT_A, child_item_data=None)
+        self.call_has_child_elements(str(ObjectId()))
+        self.check_has_child_elements_success(expected_result=True)
+
+    def test_has_child_elements_with_child_item(self):
+        """Test `has_child_elements` when there are no child system but there is a child item."""
```
rowan04, line 941:
```suggestion
        """Test `has_child_elements` when there are no child systems but there is a child item."""
```

test/unit/services/test_catalogue_item.py
```diff
@@ -1147,17 +1157,52 @@ def call_delete(self, catalogue_item_id: str) -> None:
...
 
     def test_delete(self):
         """Test deleting a catalogue item."""
 
+        self.mock_delete(has_child_elements=False)
         self.call_delete(str(ObjectId()))
         self.check_delete_success()
+
+    def test_delete_with_child_elements(self):
+        """Testing deleting a catalogue item when it has child elements."""
```
rowan04, line 1200:
```suggestion
        """Test deleting a catalogue item when it has child elements."""
```

test/unit/services/test_system.py
```diff
@@ -387,17 +397,50 @@ def call_delete(self, system_id: str) -> None:
...
 
     def test_delete(self):
         """Test deleting a system."""
 
+        self.mock_delete(has_child_elements=False)
         self.call_delete(str(ObjectId()))
         self.check_delete_success()
+
+    def test_delete_with_child_elements(self):
+        """Testing deleting a system when it has child elements."""
```
rowan04, line 440:
```suggestion
        """Test deleting a system when it has child elements."""
```

rowan04 (commented):
Code looks good! A few small comments, haven't checked functionality.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -213,10 +214,21 @@ def update(self, catalogue_item_id: str, catalogue_item: CatalogueItemPatchSchem
...
         :param catalogue_item_id: The ID of the catalogue item to delete.
+        :param access_token: The JWT access token to use for auth with the Object Storage API if object storage enabled.
         """
-        return self._catalogue_item_repository.delete(catalogue_item_id)
+        if self._catalogue_item_repository.has_child_elements(catalogue_item_id):
+            raise ChildElementsExistError(
+                f"Catalogue item with ID {catalogue_item_id} has child elements and cannot be deleted"
+            )
+
+        # First attempt to delete any attachments and/or images that might be associated with this catalogue item.
```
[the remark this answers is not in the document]
VKTB: Sorry, it is supposed to mean "we are first going to attempt to delete". I guess the way it is worded it can also mean "this is the first attempt to delete". Will a comma after "first" help or should I reword it completely?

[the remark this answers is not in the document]
rowan04: ah i see. yeah i think a comma after first would make it clear.

rowan04 (commented):
Code LGTM, but didn't test functionality!

inventory_management_system_api/core/config.py
```diff
@@ -65,6 +65,34 @@ class DatabaseConfig(BaseModel):
...
+    api_url: Optional[str] = Field(default=None, validate_default=True)
+
+    @field_validator("api_request_timeout_seconds", "api_url")
+    @classmethod
+    def validate_optional_fields(cls, field_value: str, info: ValidationInfo) -> Optional[str]:
+        """
+        Validator for the `api_request_timeout_seconds` and `api_url` fields to make them mandatory if the value of the
+        `enabled` field is `True`.
+
+        It checks if the `enabled` field has been set to `True` and raises a `TypeError` if this is the case.
```
joelvdavies, line 84: I think this line needs updating?

inventory_management_system_api/repositories/system.py
```diff
@@ -200,16 +195,16 @@ def _is_duplicate_system(
         )
         return system is not None
 
-    def _has_child_elements(self, system_id: CustomObjectId, session: Optional[ClientSession] = None) -> bool:
+    def has_child_elements(self, system_id: str, session: Optional[ClientSession] = None) -> bool:
```
joelvdavies, line 198: Is it worth doing the same to catalogue categories considering its the only one still using CustomObjectId?

joelvdavies (commented):
Two very minor comments, worked well for me.

joelvdavies (approved):

rowan04 (approved):
lgtm 😁🎉

PR #481, opened 2025-03-21
Move expected lifetime from catalogue items to items #479

### LINKED ISSUE #479: Move expected lifetime from catalogue items to items
CLF requested that it be moved to items but remain user entered. This will cover the fact it really changes per catalogue item-system pairing by forcing users to add it for every single item instead until calculating the lifetime from the history is implemented.

See corresponding front end issue here https://github.com/ral-facilities/inventory-management-system/issues/1323.

joelvdavies:
## Description

Moves expected lifetime from catalogue items to items as in #479.

As the migrations are not in production, I have modified the existing one to act as though this was always the case. As a result to test you will need to use the following

```bash
python3.12 ./scripts/dev_cli.py db-import
docker exec -it inventory_management_system_api_container ims-migrate set latest
```
(Everyone will need to do this part once this PR is merged)

Then run the backward and forward migrations through:
```bash
docker exec -it inventory_management_system_api_container ims-migrate backward 20241016101400_expected_lifetime_migration
docker exec -it inventory_management_system_api_container ims-migrate forward latest
```
The first of these effectively goes back to the state preprod/prod are in, then the second brings it back (although will wipe out any expected lifetime values populated by the mock data generation script in the process).

This will need https://github.com/ral-facilities/inventory-management-system/issues/1323 to be ready before they can be merged.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #479


VKTB (approved):

joelvdavies:
This is now no longer needed, after discussion the requested change has been reversed again.

PR #487, opened 2025-04-01
Fix `ims-migrate` migration script in production Docker image

### LINKED ISSUE #483: `ims-migrate` migration script not working
The `ims-migrate` script used for migration is not working. I am using the production Docker image from https://github.com/ral-facilities/inventory-management-system-api/pull/480 which has the latest `develop` changes.

```bash
[~]$ docker exec -it inventory_management_system_api_container ims-migrate status
OCI runtime exec failed: exec failed: unable to start container process: exec: "ims-migrate": executable file not found in $PATH: unknown
```

VKTB:
## Description
The `requirements.txt` file with all the production dependencies pinned was introduced as part of https://github.com/ral-facilities/inventory-management-system-api/pull/471 so I thought that there is no need to copy the `pyproject.toml` file and do `pip install .` anymore. However, it turns out the `ims-migrate` migration script needs to be installed so we still have to do this.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #483

### DIFF
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -5,15 +5,16 @@ WORKDIR /inventory-management-system-api-run
 # Requirement when using a different workdir to get scripts to import correctly
 ENV PYTHONPATH="${PYTHONPATH}:/inventory-management-system-api-run"
 
-COPY requirements.txt ./
+COPY pyproject.toml requirements.txt ./
 # Copy inventory_management_system_api source files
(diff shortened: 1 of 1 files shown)


joelvdavies (approved):
Yes sorry must have missed it. Its also good to have it to keep the metadata of the installed package as well.

PR #526, opened 2025-06-06
Prevent deletion of catalogue items that are replacements for others #525

### LINKED ISSUE #525: Possible to delete catalogue items that are obsolete replacement items for others
If you set Catalogue item B to be the obsolete replacement catalogue item for catalogue item A, then delete Catalogue item B, the link will be to a non-existent ID, on the front end this results in a click here link with nothing to link to, and the table/landing page also creates toasts saying `catalogue item not found`.

The simplest fix is preventing the deletion in this case. However unlike catalogue categories/items there is not a clear way to look up what is using it as a replacement, so this fix would make such entities very difficult to delete at all. Another solution down the line would be to include some sort of warning message that says its being used like this and then if still deleting - to set the obsolete replacement item id to None as part of the deletion. This solution is more complex however, and may be unlikely in the first place.

joelvdavies:
## Description

See #525. Prevents deletion of catalogue items that are assigned as the replacement of obsolete catalogue items.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #525

### DIFF
--- a/inventory_management_system_api/repositories/catalogue_item.py
+++ b/inventory_management_system_api/repositories/catalogue_item.py
@@ -137,16 +137,17 @@ class CatalogueItemRepo:
         )
         return item is not None
 
-    def is_a_replacement_for(self, catalogue_item_id: str, session: Optional[ClientSession] = None) -> bool:
+    def is_replacement_for(self, catalogue_item_id: str, session: Optional[ClientSession] = None) -> bool:
         """
-        Check if a catalogue item is an obsolete replacement catalogue item for at least one catalogue item.
+        Check if a catalogue item is the replacement for at least one obsolete catalogue item.
 
         :param catalogue_item_id: The ID of the catalogue item to check.
         :param session: PyMongo ClientSession to use for database operations.
-        :return: True if the catalogue item is a replacement for another catalogue item, False otherwise.
+        :return: True if the catalogue item is the replacement for at least one obsolete catalogue item, False
+                 otherwise.
         """
         logger.info(
-            "Checking if catalogue item with ID '%s' is a replacement for any obsolete catalogue item",
+            "Checking if catalogue item with ID '%s' is the replacement for an obsolete catalogue item",
             catalogue_item_id,
         )
         item = self._catalogue_items_collection.find_one(
--- a/inventory_management_system_api/repositories/catalogue_item.py
+++ b/inventory_management_system_api/repositories/catalogue_item.py
@@ -127,13 +127,30 @@ class CatalogueItemRepo:
 
         Child elements in this case means whether a catalogue item has child items
 
-        :param catalogue_item_id: The ID of the catalogue item to check
-        :param session: PyMongo ClientSession to use for database operations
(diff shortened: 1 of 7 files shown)


inventory_management_system_api/services/catalogue_item.py
```diff
@@ -220,12 +221,20 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
 
         :param catalogue_item_id: The ID of the catalogue item to delete.
         :param access_token: The JWT access token to use for auth with the Object Storage API if object storage enabled.
+        :raises ChildElementsExistError: If the catalogue item has child elements.
+        :raises IsAReplacementForError: If the catalogue item is a replacement for at least one other catalogue item.
```
VKTB, line 225:
```suggestion
        :raises IsAReplacementForError: If the catalogue item is the replacement for at least one obsolete catalogue
            item.
```

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -220,12 +221,20 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
...
         """
         if self._catalogue_item_repository.has_child_elements(catalogue_item_id):
             raise ChildElementsExistError(
                 f"Catalogue item with ID {catalogue_item_id} has child elements and cannot be deleted"
             )
 
+        if self._catalogue_item_repository.is_a_replacement_for(catalogue_item_id):
+            raise IsAReplacementForError(
+                f"Catalogue item with ID {catalogue_item_id} is the replacement catalogue item for at least one "
+                "catalogue item and cannot be deleted"
```
VKTB, line 235:
```suggestion
                f"Catalogue item with ID {catalogue_item_id} is the replacement for at least one obsolete catalogue "
                "item and cannot be deleted"
```

VKTB (commented):
As discussed, a few comments to improve consistency. Let me know your thoughts.

inventory_management_system_api/repositories/catalogue_item.py
```diff
@@ -137,6 +137,23 @@ def has_child_elements(self, catalogue_item_id: str, session: Optional[ClientSes
...
+        :param catalogue_item_id: The ID of the catalogue item to check.
+        :param session: PyMongo ClientSession to use for database operations.
+        :return: True if the catalogue item is a replacement for another catalogue item, False otherwise.
+        """
+        logger.info(
+            "Checking if catalogue item with ID '%s' is a replacement for any obsolete catalogue item",
+            catalogue_item_id,
+        )
+        item = self._catalogue_items_collection.find_one(
+            {"obsolete_replacement_catalogue_item_id": CustomObjectId(catalogue_item_id)}, session=session
```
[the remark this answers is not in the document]
joelvdavies: To prevent any possible database issue, no, even if it were not defined as obsolete any more if the ID is still assigned it shouldn't be removed.

VKTB (approved):

Issue #546, opened 2025-06-23
Add a settings collection with spares definition


joelvdavies:
Similar to https://github.com/ral-facilities/inventory-management-system-api/pull/421? But in this case without the put endpoint. Here need to store a list of system type ids that contain spares. This needs to be added in the mock data generation script and the mock data dump should be updated.

PR #558, opened 2025-06-27
Add system types to systems #537

### LINKED ISSUE #537: Add system type field to systems
Following on from https://github.com/ral-facilities/inventory-management-system-api/issues/536. We need to

- [x] Add a system type id field to system models and schemas
- Ensure that when creating a new system:
  - At root:
     - [x] Allow the `system_type_id` to be anything valid
   - With a parent system:
     - [x] Allow only the same `system_type_id` as its parent
- Ensure that when editing a system:
   - [x] Allow `system_type_id` to change only if the system has no child systems or items
   - At root:
     - [x] Allow the `system_type_id` to be anything valid
   - With a parent system:
     - [x] Allow only the same `system_type_id` as its parent
    - [x] Allow `parent_id` to change only if the (potentially updated) type is also the same for it
- Add a migration script ready for production (this will be specific to EPAC - but as its just EPAC should be fine at the moment, otherwise need this to be separate to the standard migrations or otherwise take in the system type values to create and how to assign them to existing systems)
    - [x] Should add the relevant system types (`storage`, `operational`, `quarantine`, `scrapped`)
    - [x] Should add the system type field for each of the root systems (matching the name - e.g. Storage -> Storage, Quarantine -> Quarantine, Scrapped -> Scrapped, Anything else -> Operational)
    - [x] Should add the system type field for each of the child systems, matching the same type as their parent

joelvdavies:
## Description

See #537. This PR
- Adds a `db-clear` command to the `dev_cli` script (Was useful for wiping my database before I started any work)
- Adds a mandatory `type_id` field to system models and schemas
- Updates the mock data & generation script to include a random selection of `type_ids` in the systems
- For creating:
  - The `type_id` can be any valid system type if the `parent_id` is None (root system)
  - The `type_id` can only be the same as the parent when `parent_id` is not None (not at root)
- For updating:
  - The `type_id` cannot be modified unless the system has no children (i.e. no subsystems or items in it)
  - The `type_id` can be any valid system type if the current or updated `parent_id` is None (at root)
  - The `type_id` can only be the same as the new or updated parent when the new or updated `parent_id` is not None (not at root)
  - The `parent_id` can only be changed when the current or updated `type_id` matches that of the new or updated parent

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

To test the migrations:

1. Checkout an older branch
2. Import the mock data with `python3.12 ./scripts/dev_cli.py db-import`
3. Add a system named Storage and one named Scrapped with some subsystems
4. Checkout this branch
5. `docker compose build` and `docker compose up`
6. Delete the `system_types` collection
7. Execute the forward migration using  `docker exec -it inventory-management-system-api ims-migrate forward latest`
8. Check the data (type_id should be added correctly to systems, these should be Storage for a system named Storage, Scrapped for Scrapped and Operational for everything else - ensure the nested systems also have the same type as their parent)
9. Execute the backward migration `docker exec -it inventory-management-system-api ims-migrate backward 20250703083422_system_types_migration`
10. Check the data (system_types collection should be deleted, and all mention of type_id removed from systems)

## Agile board tracking

Closes #537


inventory_management_system_api/services/system.py
```diff
@@ -7,43 +7,69 @@
...
+        system_repository: Annotated[SystemRepo, Depends(SystemRepo)],
+        system_type_repository: Annotated[SystemTypeRepo, Depends(SystemTypeRepo)],
+    ) -> None:
         """
-        Initialise the `SystemService` with a `SystemRepo` repository
+        Initialise the `SystemService` with a `SystemRepo` repository.
 
-        :param system_repository: `SystemRepo` repository to use
+        :param system_repository: `SystemRepo` repository to use.
+        :param system_type_repo: `SystemTypeRepo` repository to use.
```
VKTB, line 38:
```suggestion
        :param system_type_repository: `SystemTypeRepo` repository to use.
```

inventory_management_system_api/services/system.py
```diff
@@ -7,43 +7,69 @@
...
+        :raises MissingRecordError: If the system type specified by `type_id` doesn't exist.
+        :raises InvalidActionError: If the system being created has a different `type_id` to its parent.
         """
-        parent_id = system.parent_id
 
+        # If there is a parent, must use the same type as it
+        if system.parent_id is not None:
+            parent_system = self._system_repository.get(system.parent_id)
+            if not parent_system:
+                raise MissingRecordError(f"No parent system found with ID: {system.parent_id}")
```
VKTB, line 58: `create` repo method does this check too.

inventory_management_system_api/services/system.py
```diff
@@ -7,43 +7,69 @@
...
+            if system.type_id != parent_system.type_id:
+                raise InvalidActionError("Cannot use a different type_id to the parent system")
+
+        if not self._system_type_repository.get(system.type_id):
+            raise MissingRecordError(f"No system type found with ID: {system.type_id}")
```
VKTB, line 63: I am wondering if it is better for this check to happen first?

inventory_management_system_api/services/system.py
```diff
@@ -93,6 +124,39 @@ def update(self, system_id: str, system: SystemPatchSchema) -> SystemOut:
...
+                if not self._system_type_repository.get(system.type_id):
+                    raise MissingRecordError(f"No system type found with ID: {system.type_id}")
+
+            # Ensure the system type matches the parent system type if there is a parent
+            if parent_system is not None and type_id != parent_system.type_id:
+                raise InvalidActionError(
+                    "Cannot move a system into one with a different type"
+                    if parent_id_changing
+                    else "Cannot update the system's type to be different to its parent"
+                )
```
VKTB, line 158: Could you please see if some of the logic can be moved into new method(s)?

VKTB:
Just a note regarding step 7 from the test instructions: only the ID of the Operational system type is used.

inventory_management_system_api/services/system.py
joelvdavies: I have moved it to a separate function. I was going to try and create two for it, but I didn't think I could split it very nicely. E.g.
_validate_type_update
_validate_parent_update
Would still need to return the the new/current type or parent in them to do the check at the end combining both so neither would be validating them properly as they are so related.

VKTB (approved):

joelvdavies:
As a new docker image doesn't get pushed until manually requested, these changes wont break the frontend CI when merged. So I will merge them now as I need both this and ideally some changes in #566 to work on #546.

PR #566, opened 2025-07-15
Add rules collection #538

### LINKED ISSUE #538: Add a collection containing rules
These rules need to have the following fields (names are suggestions):
- `_id`
- `source_system_type_id` - Optional (None defined a creation rule)
- `destination_system_type_id` - Optional (None defines a deletion rule)
- `destination_usage_status_id` - Optional - (Only None if a deletion rule, otherwise should be defined and valid)

For development purposes we should ensure that the mock data generation script is updated to add some defaults to these. A future task will be created to consider how a sysadmin should set these up/modify them (along with the spares definition/rules).

- [ ] We must also ensure that usage statuses can't be deleted if they are used in any rules.

joelvdavies:
## Description
- Adds predefined usage statuses to the dev/test database on start up (as rules will need to know the ids in order to set them up)
  - New, In Use, Used, Scrapped
- Previously we cleared the usage status collection every test run. To ensure we can still run tests for adding usage statuses without conflict this is still required but causes a problem due to importing the predefined usage statuses with the mongodb startup script. There are a few potential solutions to this
  - We delete only the usage statuses that are not one of the predefined ones between test runs (this is the option I chose)
    -  This does mean I had to remove `test_list_no_usage_statuses` in usage status e2e tests as now there cannot be none
  - We delete all usage statuses and add the predefined usage statuses at the start of all e2e tests (Seemed excessive - especially as will need for dev purposes and would mean repeating the logic in both the conftest and start up script)
- Adds some predefined rules to the dev and test databases on start up specifically:
  - None -> Storage: New
  - Storage -> None: None
  - Storage -> Operational: In Use
  - Operational -> Storage: Used
  - Operational -> Scrapped: Scrapped
- Prevents deletion of usage statuses that are being used in rules
Note to self: Update error messages in msw handlers on frontend.

## Testing instructions

To use this PR for testing and for after this is merged and pulled for the first time you will need to delete the test database using the updated `python3.12 ./scripts/dev_cli.py db-clear` command. Otherwise using db-import should work as normal for the dev database.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #538

### DIFF
--- a/inventory_management_system_api/repositories/usage_status.py
+++ b/inventory_management_system_api/repositories/usage_status.py
@@ -87,12 +87,13 @@ class UsageStatusRepo:
         """
         Delete a usage status by its ID from a MongoDB database.
 
-        Checks if usage status is a part of an item, and does not delete if it is
+        Checks if usage status is part of an item, and does not delete if it is.
 
-        :param usage_status_id: The ID of the usage status to delete
-        :param session: PyMongo ClientSession to use for database operations
-        :raises PartOfItemError: if usage status is a part of a catalogue item
-        :raises MissingRecordError: if supplied usage status ID does not exist in the database
+        :param usage_status_id: The ID of the usage status to delete.
+        :param session: PyMongo ClientSession to use for database operations.
+        :raises PartOfItemError: If usage status is part of a catalogue item.
+        :raises PartOfItemError: If usage status is part of a rule.
+        :raises MissingRecordError: If supplied usage status ID does not exist in the database.
         """
         usage_status_id = CustomObjectId(usage_status_id)
         if self._is_usage_status_in_item(usage_status_id, session=session):


scripts/setup.mongodb
```diff
@@ -10,15 +33,65 @@ try {
...
+  // Add some rules if the collection doesn't already exist
+  if (!db.getCollectionNames().includes("rules")) {
+    console.log(`Populating rules in ${databaseName}...`);
+
+    db.rules.insertMany([
+      {
+        // Creation of items in storage -> New
+        src_system_type_id: null,
+        dst_system_type_id: SYSTEM_TYPE_ID_STORAGE,
+        usage_status_id: USAGE_STATUS_ID_NEW,
```
joelvdavies, line 69:
```suggestion
        dst_usage_status_id: USAGE_STATUS_ID_NEW,
```
Might be clearer that it relates to the value that it will be at the destination. Any thoughts on the name of these fields?

scripts/setup.mongodb
```diff
@@ -10,14 +32,64 @@ try {
...
+  // Add some rules if the collection doesn't already exist
+  if (!db.getCollectionNames().includes("rules")) {
+    console.log(`Populating rules in ${databaseName}...`);
+
+    db.rules.insertMany([
+      {
+        // Creation of items in storage -> New
+        src_system_type_id: null,
+        dst_system_type_id: SYSTEM_TYPE_ID_STORAGE,
+        dst_usage_status_id: USAGE_STATUS_ID_NEW,
```
VKTB, line 67: Is `dst_` needed before `usage_status_id`? There is no `src_usage_status_id` so I am not sure if it is needed.

VKTB: We discussed this and agreed to keep as it is.

VKTB (approved):

scripts/setup.mongodb
joelvdavies: Just repeating what I said for future reference: "Originally I just left it out, then added it thinking without it just says usage status without explaining what it means, and I felt putting dst_ there kind of suggests what its for - it defines the usage status at the destination. There isnt a src one since its irrelevant and will just be overridden."

Issue #550, opened 2025-06-23
Implement authorisation logic to usage statuses


joelvdavies:
Only an admin should be allowed to use:
- POST /usage-statuses
- DELETE /usage-statuses/{usage_status_id} (should still prevent if used somewhere - this will also include rules after https://github.com/ral-facilities/inventory-management-system-api/issues/538)

PR #602, opened 2025-09-17
Implement authorisation logic for usage statuses #550

MatteoGuarnaccia5:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

connect to #{issue number}


inventory_management_system_api/routers/v1/usage_status.py
```diff
@@ -91,8 +97,14 @@ def get_usage_status(
 def delete_usage_status(
     usage_status_id: Annotated[str, Path(description="ID of the usage status to delete")],
     usage_status_service: UsageStatusServiceDep,
+    request: Request,
 ) -> None:
     logger.info("Deleting usage status with ID: %s", usage_status_id)
+    # check user is authorised to perform operation
+    jwt_bearer = JWTBearer()
+    if not jwt_bearer.is_jwt_access_token_authorised(request.state.token):
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Not authorised to perform this operation")
```
joelvdavies, line 106: Just a suggestion of something I have seen in the FastAPI docs feel free to ignore if its not suitable https://fastapi.tiangolo.com/tutorial/security/get-current-user/#create-a-user-model. This could use a custom type PrivilegedUser that contains the roles, and gets set by a dependency running this code. So instead of adding request to the params, you would add PrivilegedUser which does the same check. Equally you could have a User that you can pass to functions that will require knowledge of the scopes but aren't immediately blocked like the rules.

MatteoGuarnaccia5: Are we planning on adding 'Users' to IMS in future? If so this is a great idea. My only issue is we stray into creating unnecessary types, classes etc, to implement essentially user's. Not sure what your thoughts are @VKTB

inventory_management_system_api/core/config.py
```diff
@@ -30,6 +30,7 @@ class AuthenticationConfig(BaseModel):
     enabled: bool
     public_key_path: Optional[str] = Field(default=None, validate_default=True)
     jwt_algorithm: Optional[str] = Field(default=None, validate_default=True)
+    privileged_roles: List[str] = Field(default=[], validate_default=True)
```
joelvdavies, line 33: This is nice as the roles are configurable and we only really have an admin/non admin role right now. Just wondering how scalable it is if we added a third. Would we then need another similar variable for the other kinds of roles e.g. operator_roles?

One other potential thought I think was raised was the idea of having some field defining what it has access to like OG has, but not defined per endpoint. E.g. we could define an `ims-admin` permission, that is used just like this `privileged_roles` is, but doesn't need to be defined in config, only in the code itself. Then it can be put in the role config you made on ldap-jwt-auth, along side the `userIsAdmin` so it can be configured. Then we would have the flexibility to create whatever roles we want, with whatever permissions we want. All we would have to do is document them in the readme e.g. if we wanted we could split access to items and systems by creating permissions for each and adding them to the role config on ldap-jwt-auth. If we are to ever do something like this, I don't think my previous suggestion would make sense anyway.

joelvdavies: I see, yeah its not something that has to be done here I don't think, I was just raising it in case it was simpler to do that now, than add this extra configuration option and then remove it again. If it has the interaction you describe it is more complicated as there is an interaction with the auth-api to check the permissions, OG puts a list of all authorized routes in the token itself so there is no need.

inventory_management_system_api/auth/jwt_bearer.py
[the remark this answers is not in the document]
MatteoGuarnaccia5: That logic makes sense to me - I did not realise you could decode it without the `PUBLIC_KEY` etc. Does `verify_exp` throw an error if it has expired?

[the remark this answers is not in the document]
joelvdavies: Yeah I forgot it could do that too, its how the frontend can decode it without the key. I believe its only needed for the signature verification. Yes it will error with `jwt.exceptions.ExpiredSignatureError` I believe looking at https://pyjwt.readthedocs.io/en/stable/api.html#jwt.exceptions.ExpiredSignatureError. I'm pretty sure the default value is True though, as otherwise we would never have to refresh the tokens and they would always work.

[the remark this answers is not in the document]
VKTB: Yeah, you could when it comes to the payload. Okay, in that case we should only check that it has not expired, so:
```python
jwt.decode(encoded, options={"verify_signature": False})
```

The default value for `verify_exp` is `True` meaning that if the token has expired then it will error with `jwt.exceptions.ExpiredSignatureError` like Joel said.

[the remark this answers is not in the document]
VKTB: > Just some thoughts on the changes in this file. Maybe the `_is_jwt_access_token_valid` here should stay as it was? Its responsibility should continue to be to check that it was signed by the right private key and algorithm, it has not expired, and the payload contains the `username` and `roles` fields. If the validation fails then the endpoint errors.

@MatteoGuarnaccia5 My suggestion was that the `_is_jwt_access_token_valid` method also checks that it contains the `roles` field. This way, we are sure that the payload has a `roles` field when decoding the token in the `is_jwt_access_token_authorised` method so there won't be a need to do any of the `.get()` dict stuff.

inventory_management_system_api/auth/jwt_bearer.py
```diff
@@ -64,3 +64,18 @@ def _is_jwt_access_token_valid(self, access_token: str) -> bool:
...
+
+        It does this by checking that the token's payload contains roles, and that (at least) one of
+        these roles is one of the configured privileged_roles
+
+        :param access_token: The JWT access token to check
+        :return `True` if the JWT access token's payload contains roles, and overlaps the configured
+        privileged_roles, `False` otherwise.
+        """
+        logging.info("Checking if JWT access token is authorised for operation")
+        payload = jwt.decode(access_token, options={"verify_signature": False})
```
VKTB, line 80: What will happen if the decode here fails? How does the endpoint behave?

inventory_management_system_api/routers/v1/usage_status.py
```diff
@@ -37,11 +39,19 @@
...
 ) -> UsageStatusSchema:
     logger.info("Creating a new usage status")
     logger.debug("Usage status data: %s", usage_status)
 
+    # check user is authorised to perform operation
+    if config.authentication.enabled is True:
+        jwt_bearer = JWTBearer()
+        if not jwt_bearer.is_jwt_access_token_authorised(request.state.token):
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED, detail="Not authorised to perform this operation"
```
VKTB, line 52: `401` is not the right code for this case. It is used when the user is not authenticated.

VKTB (changes requested):

inventory_management_system_api/auth/jwt_bearer.py
```diff
@@ -77,5 +77,10 @@ def is_jwt_access_token_authorised(self, access_token: str) -> bool:
...
         logging.info("Checking if JWT access token is authorised for operation")
-        payload = jwt.decode(access_token, options={"verify_signature": False})
-        return any(role in config.authentication.privileged_roles for role in (payload or {}).get("roles", []))
+        try:
+            payload = jwt.decode(access_token, options={"verify_signature": False})
+            return any(role in config.authentication.privileged_roles for role in payload["roles"])
+
+        except Exception:  # pylint: disable=broad-exception-caught
+            logger.exception("Error decoding JWT access token")
+            return False
```
VKTB, line 86: Shouldn't this return `False` only if it is not authorised? If the decoding fails then we are 99% sure that it is because the token has expired which is different to "you are not authorised".

inventory_management_system_api/auth/jwt_bearer.py
VKTB: See my comment https://github.com/ral-facilities/inventory-management-system-api/pull/602#discussion_r2406192598

inventory_management_system_api/auth/jwt_bearer.py
VKTB: Yeah, this is why I said "maybe we want it to verify the expiration time just in case" in my other comment and Joel mentioned that a `jwt.exceptions.ExpiredSignatureError` will be raised if it is expired. I think it should return a `401` if it has expired but I am not sure it is a good idea for the `is_jwt_access_token_authorised` method to be raising `HTTPException`. I think it should raise our own custom exception which is then caught and handled higher up.

Let me know your thoughts. Also tagging @joelvdavies.

joelvdavies: I think this is comparible to the `_is_jwt_access_token_valid` already present - it is only supposed to return if its valid or not, so it returns False in the case of an exception. Assuming the same for this case I think of the `is_jwt_access_token_authorised` as only returning if they are authorised or not. For this reason I would agree with Matteo here. However I also agree with Viktor that it should not give a 500 in this off chance and should reflect the 401 indicating it has expired correctly. I think doing this with the custom exception would make unecessarily complicated in the endpoints themselves by adding another try except around the `is_jwt_access_token_authorised`. 

As we already verify the expiry time once, perhaps in the spirit of the `_is_jwt_access_token_valid`, we ignore the  expiry time for `is_jwt_access_token_authorised` instead (using the `"verify_exp": False`) so it only checks the contents allows it to be authorised but doesnt validate the token at all? This would save the complication and it should be possible to leave the rest as is.

MatteoGuarnaccia5: I agree with Joel, the token has already been 'authenticated' so checking it's expiry whilst decoding it for an authorisation check does not make much sense - you're double authenticating/validating the token in essence - which is what we wanted to avoid doing in the first place. Hence the result of a decoding failure shouldn't be the same for authorising vs authenticating

VKTB (dismissed):

.env.example
```diff
@@ -9,6 +9,7 @@ API__ALLOWED_CORS_METHODS=["*"]
 AUTHENTICATION__ENABLED=true
 AUTHENTICATION__PUBLIC_KEY_PATH=./keys/jwt-key.pub
 AUTHENTICATION__JWT_ALGORITHM=RS256
+AUTHENTICATION__PRIVILEGED_ROLES=["admin"] # Will become more complex in future
```
joelvdavies:
```suggestion
AUTHENTICATION__PRIVILEGED_ROLES=["admin"]
```
I think its fine without the comment, unless there is any urgency on extra features it could stay like this for a long time.

inventory_management_system_api/auth/authorisation.py
```diff
@@ -0,0 +1,18 @@
...
+from fastapi import Depends, Request
+from inventory_management_system_api.auth.jwt_bearer import JWTBearer
+from inventory_management_system_api.core.config import config
+
+
+def _authorised_dep(request: Request) -> bool:
+    if config.authentication.enabled is True:
+        jwt_bearer = JWTBearer()
+        return jwt_bearer.is_jwt_access_token_authorised(request.state.token)
+    return True
```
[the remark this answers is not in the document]
joelvdavies: I think technically we can't guarentee the order. But would that be a problem? Its still doing the both checks eitherway and if it effects this, then it probably means the same with service dependencies. Putting the authentication in the middleware is the other way to ensure it happens before as had to be done for the object storage api to prevent it with the file uploads as it would allow the whole file to upload before authentication as it is also a dependency.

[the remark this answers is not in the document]
VKTB: From some resources I came across online, the order in which the dependencies are run may not be guaranteed.

> But would that be a problem?

I don't know. Walking through a case may help.
1. `is_jwt_access_token_authorised` dependency runs first and it returns `True`. This means the user is authorised but the token's signature and expiry were not verified.
2. `JWTBearer` dependency runs after the `is_jwt_access_token_authorised` dependency returns.
    - `is_jwt_access_token_authorised` returns `True` (token verification passes) so the endpoint will proceed with executing the request. (**This is what we want to happen so it would not be a problem**)
    - `is_jwt_access_token_authorised` returns `False` (token verification fails) so the endpoint will not proceed with executing the request and return `403` to the client. (**This is what we want to happen so it would not be a problem**)

The above assumes that any of the logic in the endpoint body methods will execute **ONLY** after the dependencies fully complete/return. 

However, I don't know what the behaviour will be if things are asynchronous i.e. will the logic in the endpoint body methods begin to execute if dependencies have not completed/returned. In the case of this API, we don't have `async` endpoints (so that is one less thing to worry about) **but** [the callable method of the `JWTBearer` class](https://github.com/ral-facilities/inventory-management-system-api/blob/develop/inventory_management_system_api/auth/jwt_bearer.py#L31) is `async` so I don't know how that is going to behave.

---

> Isn't it guaranteed to run as it is hard coded into the specific endpoints? Or do dependencies not work like that?

@MatteoGuarnaccia5 They are guaranteed to run but the order in which they are run may not be guaranteed. I also don't know if they will all run before any logic in the endpoint body methods gets executed.

inventory_management_system_api/routers/v1/usage_status.py
```diff
@@ -91,8 +96,14 @@ def get_usage_status(
 def delete_usage_status(
     usage_status_id: Annotated[str, Path(description="ID of the usage status to delete")],
     usage_status_service: UsageStatusServiceDep,
+    authorised: AuthorisedDep,
```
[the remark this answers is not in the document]
joelvdavies: Yes, they should be.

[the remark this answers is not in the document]
joelvdavies: Just spoken to Joshua, its fine for the front end, stick with the latest comment on isPrivilegedUser.

joelvdavies (dismissed):

joelvdavies (approved):

PR #603, opened 2025-09-17
Configure system types #601

### LINKED ISSUE #601: Add way to configure system types on CLI
Following on from #600. We need a way to configure the system types available on the system.

joelvdavies:
## Description

Adds commands for configuring system types. Each command is interactive so requires further input once called.

### Creating a system type
```
ims create system-type
```
image

### Updating a system type
```
ims update system-type
```
image

### Deleting a system type
```
ims delete system-type
```
image

This also checks that the type is not in use before deleting. As it is possible for another user to create a new system using an existing type after this check but before deletion, a warning message is printed out similar to the spares one asking that it be run when no one else is using the system.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test trying to delete system types that are being used (by a system, by rules and by the spares definition)

## Agile board tracking

Closes #601

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,6 +399,42 @@ or in Docker
 docker exec -it inventory-management-system-api ims --help
 ```
 
+#### Managing system types
+
+System types are used to distinguish different types of systems from each other and are required to exist before any systems can be created. They are also used in the spares definition and rules.
+
+System types can be created, updated and deleted using the following interactive commands:
+
+```bash
+ims create system-type
+ims update system-type
+ims delete system-type
+```
+
+Running each of these will guide you through the process.
+
+#### Configuring the spares definition
+
+The spares definition is a list of system types that define which systems contain spares. E.g. If you have system type
+of `Storage`, and define the spares definition as `Storage`. Then a spare of a given catalogue item will be defined as
+any item that is within a system of with a type of `Storage`.
+
+When first setup, IMS does not have a spares definition assigned and so will not attempt to calculate the number of
+spares you have for each catalogue item. To configure the spares definition and enable this functionality use
+
+```bash
+ims configure spares-definition
+```
+
+and follow its instructions. This will calculate and update the `number_of_spares` field on all catalogue items and may
+also be used to change the spares definition once already setup.
+
+NOTE: Please ensure that no one is using ims-api when executing this. Otherwise it is possible to miscount the number of
+spares. If using a reverse proxy to provide access, we recommend first shutting it down before doing this.
+
+Subsequently, whenever a new item is added, moved or deleted, the number of spares inside the effected catalogue item
+will be recalculated accordingly
+
 #### Migrations
 
 ##### Adding a migration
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -399,6 +399,42 @@ or in Docker
 docker exec -it inventory-management-system-api ims --help
 ```
 
+#### Managing system types
+
+System types are used to distinguish different types of systems from each other and are required to exist before any systems can be created. They are also used in the spares definition and rules.
```
VKTB, line 404:
```suggestion
System types are used to distinguish different types of systems from each other and are required to exist before any
systems can be created. They are also used in the spares definition and rules.
```

inventory_management_system_api/cli/create.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # Display a table of existing system types for reference
+    current_system_types = system_type_service.list()
+    console.print("Below is the current list of system types available:")
+    display_indexed_system_types(current_system_types)
+
+    # Obtain name of the new system type and ensure it doesn't already exist (case insensitive)
+    new_system_type_value = Prompt.ask("Please enter the value of the new system type")
+    if system_types_collection.find_one({"value": re.compile(new_system_type_value, re.IGNORECASE)}):
+        sys.exit("Already exists!")
```
VKTB, line 34: I would change this to:
```suggestion
        sys.exit("A system type with the same value already exists!")
```
Maybe also make the text colour red if that is possible?

inventory_management_system_api/cli/delete.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Obtain the current spares definition
+    # pylint:disable=fixme
+    # TODO: Obtain from the setting service directly rather than via the repo once implemented in #549
+    # pylint:disable=protected-access
+    current_spares_definition = setting_service._setting_repository.get(SparesDefinitionOut)
+
+    # Ensure the system type is not used in any system, rule or the spares definition
+    if (
+        systems_collection.find_one({"system_type_id": selected_type_id})
```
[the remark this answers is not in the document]
joelvdavies: 🤦‍♂️

inventory_management_system_api/cli/create.py
joelvdavies: Here I definitely can, unfortunately typer/rich don't have an exit that will take formatting. So I didn't bother initially due to the migrations which currently have some unit tests relying on the messages from the SystemExit exception raised to know what happened. To make this work I would need to do
```python
console.print("[red]A system type with the same value already exists![/]")
typer.Exit(1)
```
I will see if I can come up with something that will work for both.

joelvdavies: I have currently created a generic `MigrationError` to use in the migrations code, then caught it in the CLI to handle the printing/exit so all exit messages are now printed in red.

joelvdavies:
> Updating the value of an existing system type to one that already exists succeeds.
> 
> ```shell
> $ ims update system-type
> Below is the current list of system types available:
> ┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┓
> ┃ Index ┃ ID                       ┃ Value       ┃
> ┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━┩
> │ 1     │ 685e5dce6e347e39d459c5ea │ Storage     │
> │ 2     │ 685e5dce6e347e39d459c5eb │ Operational │
> │ 3     │ 685e5dce6e347e39d459c5ec │ Scrapped    │
> └───────┴──────────────────────────┴─────────────┘
> 
> Please enter the index of the system type to update: 3
> Please enter the new value of the selected system type: Storage
> You are updating: 3 (Scrapped -> Storage)
> 
> Are you sure you want to apply this change? [y/N]: y
> 
> Success! 🎉
> ```

Thanks, good spot. I have now prevented this using a similar check to our repos, it prevents changing the value to match another (case insensitive) but still allows the system type being edited to have its case changed e.g. `storage -> Storage`.

VKTB (approved):

PR #611, opened 2025-09-26
Configure rules #604

### LINKED ISSUE #604: Add way to configure rules on CLI
Following on from #603. We need a way to configure the rules available on the system.

joelvdavies:
## Description

Adds ability to configure the rules via the admin cli. Adds a way to create new rules and delete old ones. Update is left out as would be much more complicated and can be just as easily accomplished with deleting the old one and creating a new one.

See readme for instructions. Adds two commands `ims create rule` and `ims delete rule`.

e.g. Creating a new moving rule:
image

Deleting a rule:
image

### Notes
- Updated typer to newer version so could use a set of choices for the `ims create rule` rule_type parameter. Due to this you will need to either rebuild the container if using docker exec (via `docker compose build --no-cache`) or ensure your venv is updated using `python -m pip install -r requirements.txt` from within it.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test creating duplicate rules (should be prevented)

## Agile board tracking

Closes #604

### DIFF
--- a/inventory_management_system_api/cli/create.py
+++ b/inventory_management_system_api/cli/create.py
@@ -118,7 +118,10 @@ def get_user_constructed_rule(
 
 
 def display_user_constructed_rule(
-    rule_type: RuleType, src_system_type: SystemTypeOut, dst_system_type: SystemTypeOut, dst_usage_status: SystemTypeOut
+    rule_type: RuleType,
+    src_system_type: SystemTypeOut,
+    dst_system_type: SystemTypeOut,
+    dst_usage_status: UsageStatusOut,
 ):
     """Displays a user constructed rule obtained from `get_user_constructed_rule`."""
 
--- a/README.md
+++ b/README.md
@@ -416,7 +416,7 @@ Running each of these will guide you through the process.
 
 #### Managing rules
 
-Rules define what is operations are possible for items in relation to systems. They are split into three categories:
+Rules define what operations are possible for items in relation to systems. They are split into three categories:
 
 - Creation: These define which types of system's items can be created in, and what usage status they should have.
 - Deletion: These define which types of system's items can be deleted from (They have no restriction on usage status)


inventory_management_system_api/cli/create.py
```diff
@@ -30,8 +48,137 @@ def system_type():
...
+    table.add_row(
+        (f"{src_system_type.id} [orange1]({src_system_type.value})[/]" if src_system_type else "None"),
+        (f"{dst_system_type.id} [orange1]({dst_system_type.value})[/]" if dst_system_type else "None"),
+        (f"{dst_usage_status.id} [orange1]({dst_usage_status.value})[/]" if dst_usage_status else "None"),
+    )
+    console.print(table)
+    console.print()
+
+    # Output an explanation of the rule
+    display_rule_explanation(src_system_type, dst_system_type, dst_usage_status)
```
VKTB, line 137: I get the following warning for `dst_usage_status`: `Expected type 'UsageStatusOut | None', got 'SystemTypeOut' instead `

inventory_management_system_api/cli/create.py
```diff
@@ -30,8 +48,137 @@ def system_type():
...
+    rule_data = {
+        "src_system_type_id": CustomObjectId(src_system_type.id) if src_system_type else None,
+        "dst_system_type_id": CustomObjectId(dst_system_type.id) if dst_system_type else None,
+        "dst_usage_status_id": CustomObjectId(dst_usage_status.id) if dst_usage_status else None,
+    }
+    if rules_collection.find_one(rule_data):
+        exit_with_error("The selected rule already exists!")
+
+    # Output the user selected rule and request confirmation before adding it
+    display_user_constructed_rule(rule_type, src_system_type, dst_system_type, dst_usage_status)
```
VKTB, line 175: I get the following warning for `dst_usage_status`: `Expected type 'SystemTypeOut', got 'UsageStatusOut | None' instead`

VKTB (approved):

PR #614, opened 2025-09-29
Implement item creation rules logic

### LINKED ISSUE #539: Implement item creation rules
After https://github.com/ral-facilities/inventory-management-system-api/issues/538. This should prevent an item being created with a `system_id`, unless there is a creation rule in existence for the system's type, and matching the chosen item usage status.

| intitial_system_type | final_system_type | final_usage_status |
| -------- | ------- | ------- |
| null | storage | new |

Allows creation of an item in the storage system provided the usage status of the item is new. (There could be multiple rules with different usage statuses)

VKTB:
## Description
The changes in this PR do not allow items to be created in a system with a usage status for which a creation rule does not exist.

For example, if there is creation rule that allows items to only be created in systems which are of type storage and requires their usage status to be new (see rule below), then it will not allow items to be created anywhere else and/or with a different usage status.

| `src_system_type_id` | `dst_system_type_id` | `dst_usage_status_id` |
| -------- | ------- | ------- |
| `null` | `685e5dce6e347e39d459c5ea` | `6874cf5dee233ec6441860a0` |

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes #539

### DIFF
--- a/test/mock_data.py
+++ b/test/mock_data.py
@@ -991,6 +991,13 @@ RULE_GET_DATA_STORAGE_CREATION = {
     "dst_usage_status": USAGE_STATUS_GET_DATA_NEW,
 }
 
+RULE_DOCUMENT_DATA_STORAGE_CREATION = {
+    "_id": PREDEFINED_RULE_IDS[0],
+    "src_system_type_id": None,
+    "dst_system_type_id": SYSTEM_TYPE_OUT_DATA_STORAGE["_id"],
+    "dst_usage_status_id": USAGE_STATUS_OUT_DATA_NEW["_id"],
+}
+
 # Deletion from storage
 RULE_OUT_DATA_STORAGE_DELETION = {
     "_id": PREDEFINED_RULE_IDS[1],
@@ -1001,6 +1008,13 @@ RULE_OUT_DATA_STORAGE_DELETION = {
 
 RULE_GET_DATA_STORAGE_DELETION = {
     **RuleOut(**RULE_OUT_DATA_STORAGE_DELETION).model_dump(),
+}
+
+RULE_DOCUMENT_DATA_STORAGE_DELETION = {
+    "_id": PREDEFINED_RULE_IDS[1],
+    "src_system_type_id": SYSTEM_TYPE_OUT_DATA_STORAGE["_id"],
+    "dst_system_type_id": None,
+    "dst_usage_status_id": None,
 }
 
 # Storage to operational
@@ -1016,6 +1030,13 @@ RULE_GET_DATA_STORAGE_TO_OPERATIONAL = {
     "dst_usage_status": USAGE_STATUS_GET_DATA_IN_USE,
 }
 
+RULE_DOCUMENT_DATA_STORAGE_TO_OPERATIONAL = {
+    "_id": PREDEFINED_RULE_IDS[2],
+    "src_system_type_id": SYSTEM_TYPE_OUT_DATA_STORAGE["_id"],
+    "dst_system_type_id": SYSTEM_TYPE_OUT_DATA_OPERATIONAL["_id"],
+    "dst_usage_status_id": USAGE_STATUS_OUT_DATA_IN_USE["_id"],
+}
+
 # Operational to storage
 RULE_OUT_DATA_OPERATIONAL_TO_STORAGE = {
     "_id": PREDEFINED_RULE_IDS[3],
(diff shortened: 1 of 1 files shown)


VKTB:
@joelvdavies Let me know your initial thoughts. I want to get the logic right before I start looking at the tests. Thanks.

joelvdavies:
> @joelvdavies Let me know your initial thoughts. I want to get the logic right before I start looking at the tests. Thanks.

@VKTB Yes this looks good to me

inventory_management_system_api/services/item.py
```diff
@@ -102,11 +105,25 @@ def create(self, item: ItemPostSchema) -> ItemOut:
...
             raise MissingRecordError(f"No usage status found with ID: {usage_status_id}")
 
+        if not self._rule_repository.check_exists(
+            src_system_type_id=None,
+            dst_system_type_id=system.type_id,
+            dst_usage_status_id=usage_status_id,
+        ):
+            raise InvalidActionError(
+                "No rule found for creating items in the specified system with the specified usage status"
+            )
```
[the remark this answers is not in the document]
joelvdavies: Just double checked if you force it by deleting a rule after filling out the create item form it does show the type, but the message to me should be clear enough for a user:
image

test/e2e/test_setting.py
```diff
@@ -196,6 +219,14 @@ def test_create_item_after_spares_definition_set(self):
...
         """Test creating an item after the spares definition is set to one with multiple system type IDs in it.
         (This test is not repeated for update/delete as the logic used is the same)"""
+        # Add a custom rule to allow items to be created in Operational systems
+        custom_creation_rule_id = self.add_rule(
+            {
+                "src_system_type_id": None,
+                "dst_system_type_id": SYSTEM_TYPE_OUT_DATA_OPERATIONAL["_id"],
+                "dst_usage_status_id": USAGE_STATUS_OUT_DATA_NEW["_id"],
+            }
+        )
```
joelvdavies, line 229:
```suggestion
        custom_creation_rule_id = self.add_rule(RULE_DOCUMENT_DATA_OPERATIONAL_CREATION)
```
Would mean it can be reused below if its added to the mock_data file, name matches that from the function its used in.

test/mock_data.py
```diff
@@ -674,7 +674,7 @@
 BASE_CATALOGUE_ITEM_DATA_WITH_PROPERTIES = CATALOGUE_ITEM_DATA_WITH_ALL_PROPERTIES
 
 
-# New, Required values only
+# New, All required values except ids that are not predefined
```
joelvdavies, line 677:
```suggestion
# New, Required values only
```
Could you add a note at the top of the file instead, it can apply to all of them I think. Perhaps:

```
_REQUIRED_VALUES_ONLY - Is used to indicate the values given are only the required ones, except ids that are not predefined and will be added later when they are known in tests.
```

test/e2e/conftest.py
```diff
@@ -40,6 +47,14 @@ def fixture_cleanup_database_collections():
     database.settings.delete_many({})
     # Remove extra usage statuses but leave predefined ones
     database.usage_statuses.delete_many({"_id": {"$nin": PREDEFINED_USAGE_STATUS_IDS}})
+    # Ensure predefined usage statuses have not been accidentally removed and insert them if they have
+    for usage_status in [
+        USAGE_STATUS_OUT_DATA_NEW,
+        USAGE_STATUS_OUT_DATA_IN_USE,
+        USAGE_STATUS_OUT_DATA_USED,
+        USAGE_STATUS_OUT_DATA_SCRAPPED,
+    ]:
```
joelvdavies, line 56:
```suggestion
    for usage_status in PREDEFINED_USAGE_STATUS_IDS:
```
Maybe to match the above?

test/e2e/test_setting.py
```diff
@@ -212,8 +243,19 @@ def test_create_item_after_spares_definition_set_with_multiple_types(self):
         self.post_item_in_system_with_type_id(SYSTEM_TYPE_GET_DATA_STORAGE["id"])
         self.check_catalogue_item_spares([1, 3])
 
+        # Delete the rule that allows items to be created in Operational systems to avoid conflicts with other tests
+        self.delete_rule(custom_creation_rule_id)
```
joelvdavies, line 247: Is it not worth doing the same thing in the conf test to remove predefined rules like the usage status one? Right now if the test fails it wont reset and it will be stuck there. I think its just the setup.mongodb would need predefined ids to do this.

joelvdavies (commented):
Thanks for this, turned out more involved than I thought 😅. Just some suggestions on tests, worked well for me.

test/e2e/conftest.py
VKTB: But `PREDEFINED_USAGE_STATUS_IDS` is a list of IDs and this the usage statuses to insert them if the IDs do not exist.

joelvdavies: Yeah it would, its only the fact it doesnt attempt to insert it unless it doesnt exist, it is still however an update replacing the contents if it does. In my first part yeah I should have renamed the variable to `PREDEFINED_USAGE_STATUSES` which just contains the same things (just thought it might be nice to keep the definition of which ones were predefined in the mock data file).

Yeah we could just empty them at that point, then again we would still save repeatedly clearing and inserting all of them which is two separate operations. In this case you could even use upsert with update_many to do all at once.

VKTB: Looks like the only reason it modifies them is because the `created_time` and `modified_time` values differ each time these constants are referenced. At the moment we have 6 different operations - one for `delete_many` and `update_one` for each usage status. This will reduce down to 2 operations - one for `delete_many` and another for `insert_many`

joelvdavies (commented):
Just the one thing, thanks for sorting that out, this issue turned out more complicated than I expected 😄

Issue #563, opened 2025-07-11
Add `description` field to system types


joelvdavies:
For the purposes of knowing what a system type means and how its used. (https://github.com/ral-facilities/inventory-management-system-api/pull/557#issuecomment-3052442911)

PR #631, opened 2025-10-15
Add description field to system types #563

lottieo:
## Description

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

connect to #563


inventory_management_system_api/cli/create.py
```diff
@@ -50,8 +50,11 @@ def system_type():
     if system_types_collection.find_one({"value": re.compile(new_system_type_value, re.IGNORECASE)}):
         exit_with_error("A system type with the same value already exists!")
 
+    # Obtain description of the new system type
+    new_system_type_description = Prompt.ask("Please enter the description of the new system type")
+
     # Insert the new system type
-    system_types_collection.insert_one({"value": new_system_type_value})
+    system_types_collection.insert_one({"value": new_system_type_value, "description": new_system_type_description})
```
joelvdavies: Looking at the database it stores this as an empty string when you dont supply it.

Image

This should really be stored as `null` instead. To do this you would need to pass `None` here.

scripts/setup.mongodb
```diff
@@ -55,9 +55,9 @@ try {
     console.log(`Populating system types in ${databaseName}...`);
 
     db.system_types.insertMany([
-      { _id: SYSTEM_TYPE_ID_STORAGE, value: "Storage" },
-      { _id: SYSTEM_TYPE_ID_OPERATIONAL, value: "Operational" },
-      { _id: SYSTEM_TYPE_ID_SCRAPPED, value: "Scrapped" },
+      { _id: SYSTEM_TYPE_ID_STORAGE, value: "Storage", description: "storage system type" },
+      { _id: SYSTEM_TYPE_ID_OPERATIONAL, value: "Operational", description: "operational system type" },
+      { _id: SYSTEM_TYPE_ID_SCRAPPED, value: "Scrapped", description: "scrapped system type" },
```
joelvdavies:
```suggestion
      { _id: SYSTEM_TYPE_ID_STORAGE, value: "Storage", description: "Storage system type" },
      { _id: SYSTEM_TYPE_ID_OPERATIONAL, value: "Operational", description: "Operational system type" },
      { _id: SYSTEM_TYPE_ID_SCRAPPED, value: "Scrapped", description: "Scrapped system type" },
```
Although I dont think what we use really matters right now anyway, I spoke with Joshua and they could be more like "Contains items that are in storage", or "An operational system, items inside are classified as in use" or something perhaps as a hover tooltip on the frontend. It will probably be clearer down the line when the frotend issue is done. We also don't know what CLF would want to use.

VKTB (approved):

PR #659, opened 2025-11-27
Allow admins to edit catalogue category property unit #620

### LINKED ISSUE #620: Editing of unit of a property in a catalogue category by an admin
Requested by Maks from users. The ability to edit a unit of a property after it has been created at the catalogue category level. Should be admin only, and will at present require a migration to properly propagate this down through any catalogue items/items as needed. Related to frontend issue here.

MatteoGuarnaccia5:
## Description

see #620 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #620

### DIFF
--- a/inventory_management_system_api/services/catalogue_category_property.py
+++ b/inventory_management_system_api/services/catalogue_category_property.py
@@ -235,7 +235,7 @@ class CatalogueCategoryPropertyService:
             else:
                 unit = self._unit_repository.get(unit_id)
                 if not unit:
-                    raise MissingRecordError(f"No unit found with ID: {unit_id}")
+                    raise MissingRecordError(f"No unit found with ID '{unit_id}'")
                 update_data["unit"] = unit.value
 
         CatalogueCategoryPostPropertySchema.check_valid_allowed_values(
--- a/test/unit/services/test_catalogue_category_property.py
+++ b/test/unit/services/test_catalogue_category_property.py
@@ -15,6 +15,7 @@ from test.mock_data import (
     CATALOGUE_CATEGORY_PROPERTY_DATA_NUMBER_NON_MANDATORY_WITH_ALLOWED_VALUES_LIST,
     CATALOGUE_CATEGORY_PROPERTY_DATA_NUMBER_NON_MANDATORY_WITH_MM_UNIT,
     CATALOGUE_CATEGORY_PROPERTY_IN_DATA_NUMBER_NON_MANDATORY,
+    CATALOGUE_CATEGORY_PROPERTY_IN_DATA_NUMBER_NON_MANDATORY_WITH_MM_UNIT,
     UNIT_IN_DATA_MM,
 )
 from test.unit.services.conftest import BaseCatalogueServiceDSL, ServiceTestHelpers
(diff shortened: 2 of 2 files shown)


inventory_management_system_api/repositories/catalogue_item.py
```diff
@@ -210,32 +210,43 @@ def insert_property_to_all_matching(
             session=session,
         )
 
-    def update_names_of_all_properties_with_id(
-        self, property_id: str, new_property_name: str, session: Optional[ClientSession] = None
+    def update_names_and_units_of_all_properties_with_id(
```
joelvdavies: I personally feel this might be better off as a separate function. The name implies both are updated when they are not, its conditional, equally we want to keep things separate when they do different unrelated things, though I can understand wanting to avoid duplication for this. Would be the same as in item.py.

joelvdavies (commented):
Just testing and found that if I as a non admin do a post request with
{
  "unit_id": null
}
Whent the current value is also none, it succeeds and returns the property. This would ideally also fail as its attempting to change it even if it wont be any different.

test/e2e/test_catalogue_category_property.py
```diff
@@ -615,6 +621,64 @@ def test_partial_update_name_to_duplicate(self):
...
+    def test_partial_update_invalid_unit_id(self):
+        """Test updating the unit of a property to an invalid unit ID"""
+
+        property_id = self.post_test_property_and_prerequisites(CATALOGUE_CATEGORY_PROPERTY_DATA_NUMBER_NON_MANDATORY)
+
+        self.patch_property(property_id, {"unit_id": "invalid"}, use_admin_token=True)
+
+        self.check_patch_property_failed_with_detail(404, "The specified unit does not exist")
+
+    def test_partial_update_unit_not_authorised(self):
```
joelvdavies: Is it worth having the same test when changing it to itself? Just to prevent any regression in future.

test/unit/repositories/test_item.py
```diff
@@ -542,49 +542,60 @@ def test_insert_property_to_all_matching(self):
...
             },
-            array_filters=[{"elem._id": CustomObjectId(self._update_names_of_all_properties_with_id_property_id)}],
+            array_filters=[{"elem._id": CustomObjectId(self._update_all_properties_with_id_property_id)}],
             session=self.mock_session,
         )
 
 
-class TestUpdateNamesOfAllPropertiesWithID(UpdateNamesOfAllPropertiesWithIDDSL):
-    """Tests for `update_names_of_all_properties_with_id`."""
+class TestUpdateManyPropertyFieldsWithID(UpdateAllPropertiesWithIDDSL):
```
joelvdavies:
```suggestion
class TestUpdateAllPropertiesWithID(UpdateAllPropertiesWithIDDSL):
```

joelvdavies (commented):
Couple minor things as I went throught the tests. Previous issue is now fixed 🥳

VKTB (changes requested):
I have a bit more left to review but submitting this before I restart my machine.

VKTB (approved):

PR #738, opened 2026-03-25
Add last modified by field #710

### LINKED ISSUE #710: add last modified by field to all data
As part of work for history, add a `last_modified_by` field to all schemas, much like `last_modified_time`.

This will potentially require a migration of the current data.

MatteoGuarnaccia5:
## Description

Similar to #714 , implement a last modified by field for every update and create request to the main collections (excl rules, system types).

Includes migration script as well

The username is stored in the request state when the authentication check is carried out.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

closes #710


VKTB:
Could you please provide a description about the changes as I am struggling to understand the idea behind some of this work? I am not sure I understand why a `username` claim would be missing in the JWT payload.

MatteoGuarnaccia5:
> Could you please provide a description about the changes as I am struggling to understand the idea behind some of this work? I am not sure I understand why a `username` claim would be missing in the JWT payload.

Note - Discussed during sprint meeting and confusion resolved

inventory_management_system_api/migrations/scripts/20260325102658_modified_by_field.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+# pylint:disable=too-many-instance-attributes
+class Migration(BaseMigration):
+    """Migration that Add the modified_by field to all documents"""
+
+    description = "Add the modified_by field to all documents"
+
+    def __init__(self, database: Database):
+        self._catalouge_categories_collection: Collection = database.catalogue_categories
```
VKTB:
```suggestion
        self._catalogue_categories_collection: Collection = database.catalogue_categories
```

inventory_management_system_api/migrations/scripts/20260325102658_modified_by_field.py
```diff
@@ -0,0 +1,51 @@
...
+        self._catalogue_items_collection: Collection = database.catalogue_items
+        self._items_collection: Collection = database.items
+        self._manufacturers_collection: Collection = database.manufacturers
+        self._systems_collection: Collection = database.systems
+        self._units_collection: Collection = database.units
+        self._usage_statuses_collection: Collection = database.usage_statuses
+
+        self._collections = [
+            self._catalouge_categories_collection,
```
VKTB:
```suggestion
            self._catalogue_categories_collection,
```

inventory_management_system_api/migrations/scripts/20260325102658_modified_by_field.py
```diff
@@ -0,0 +1,51 @@
...
+            self._systems_collection,
+            self._units_collection,
+            self._usage_statuses_collection,
+        ]
+
+    def forward(self, session: ClientSession):
+        """Applies database changes."""
+
+        for collection in self._collections:
+            collection.update_many({}, {"$set": {"modified_by": "Development user"}}, session=session)
```
VKTB: Regarding your https://github.com/ral-facilities/inventory-management-system-api/pull/738#issuecomment-4125420042, is there are a reason why we can't just stick with one for everywhere? If we go with "IMS Admin", would that be meaningful enough for frontend users to work out that the modification has happened as a result of data migration? Or do we need to be more specific and go with something like "IMS Migration Service"?

VKTB (commented):
Related to https://github.com/ral-facilities/inventory-management-system-api/issues/737 but it's worth thinking about whether it should be possible to `PATCH` just the `modified_comment` itself. At the moment that is possible even if I am not including any other fields in the `PATCH` request.

```bash
curl -X 'PATCH' \
  'http://localhost:8000/v1/catalogue-categories/6a0afc423ebf371da7e086c8' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "modified_comment": "I am changing the modification comment"
}'
```

inventory_management_system_api/migrations/scripts/20260325102658_modified_by_field.py
MatteoGuarnaccia5: I was thinking something more like IMS migration service as IMS admin could imply an admin user - not us the developers. Its just agreeing on a name basically and adding that to the migration script instructions so it always gets used

VKTB: > as IMS admin could imply an admin user

Exactly what I thought as well. @joelvdavies @joshdimanteto thoughts on what name to use?

joelvdavies: I thought this is just for the one time script - I.e. right now people have been modifying data but it hasnt recorded who has done the past edits as it wasnt implemented, so this migration is then adding a field which would normally be the person who made the last modification, in this case I think "Unknown" is the most suitable. Its not known who the last person who modified the data was. Or are we saying that future migrations are supposed to be recorded in the history itself?

joshdimanteto: These auto-generated names will be parsed into user-friendly names in the frontend, so it’s best to keep them relevant for programmatic use when using the API directly, ensuring they provide clear context about backend services

VKTB: My understanding before you comment @joelvdavies was that data migrations (adding/removing fields) are considered modifications but you are making a good point.

> Or are we saying that future migrations are supposed to be recorded in the history itself?

I actually didn't think about this. @MatteoGuarnaccia5 Have you given this any thought? How would history work when a data migration is performed?

VKTB:
The `uv run ./scripts/dev_cli.py db-restore mock_data.dump` command from the README for populating the database fails when I ran it with the changes on this branch.

inventory_management_system_api/migrations/scripts/20260325102658_modified_by_field.py
MatteoGuarnaccia5: I thought we intended to turn off history during migrations? If that is the case then Joel's comment is completely correct and it was just a point for this specific migration (which now that my memory is coming back to me was the point of my question)

If we wish to keep history on during migrations - it will work completely fine, but I suggest having 'default' `last_modified_by` and `modified_comment` fields that should be used, so that users can see when viewing history why this random new field has been added. The defaults could be, as previously said, "IMS Migration Service" + "Migration performed by IMS development team"

inventory_management_system_api/repositories/catalogue_item.py
```diff
@@ -223,13 +234,16 @@ def update_all_properties_with_id(
...
+        :param username: The user updating the property
         :param session: PyMongo ClientSession to use for database operations
         """
 
         logger.info("Updating all properties with ID '%s' inside catalogue items in the database", property_id)
 
         set_body = {f"properties.$[elem].{k}": v for k, v in update_body.items()}
         set_body["modified_time"] = datetime.now(timezone.utc)
+        set_body["modified_by"] = username
+        set_body["modified_comment"] = "Property updated"
```
joelvdavies, line 246: I see the creation has the name in it, should the update do the same? E.g. when I create a property called New I get a modified comment of `Property 'New' created`, but if I update it its just `Property updated`.

inventory_management_system_api/routers/v1/system.py
```diff
@@ -42,11 +42,11 @@
     response_description="The created system",
     status_code=status.HTTP_201_CREATED,
 )
-def create_system(system: SystemPostSchema, system_service: SystemServiceDep) -> SystemSchema:
+def create_system(request: Request, system: SystemPostSchema, system_service: SystemServiceDep) -> SystemSchema:
     logger.info("Creating a new system")
     logger.debug("System data: %s", system)
     try:
-        system = system_service.create(system)
+        system = system_service.create(system, request.state.username)
```
joelvdavies, line 49: I would personally rather we passed around some sort of 'User' object with the username & role so we can use the same whenever needed and store that in the request state, similar to what I suggested https://github.com/ral-facilities/inventory-management-system-api/pull/602#discussion_r2358176410. But its just a suggestion.

inventory_management_system_api/repositories/catalogue_item.py
joelvdavies: Also `delete_properties` doesnt have one. Is there a reason one has it and one doesnt? Spares doesnt either which I expected when first looking at it as its an internal mechanism but now that I think of it it will be in the history too?

inventory_management_system_api/routers/v1/system.py
MatteoGuarnaccia5: I think it is quite an involved change - as it means refactoring the auth logic to suit the state, which could be beneficial, but probably best for a different PR

VKTB: @MatteoGuarnaccia5 Could you please create an issue for this to track it? We will get to it at some point and decide if we should implement it. Thanks.

MatteoGuarnaccia5:
> Related to #737 but it's worth thinking about whether it should be possible to `PATCH` just the `modified_comment` itself. At the moment that is possible even if I am not including any other fields in the `PATCH` request.
> 
> ```shell
> curl -X 'PATCH' \
>   'http://localhost:8000/v1/catalogue-categories/6a0afc423ebf371da7e086c8' \
>   -H 'accept: application/json' \
>   -H 'Content-Type: application/json' \
>   -d '{
>   "modified_comment": "I am changing the modification comment"
> }'
> ```

Post in person discussion - History handles just the modified comment field fine, it will just show that change in the frontend, given requests made via the frontend will always have data with the comment field the only case this occurs is if someone purposely patches just the comment. If CLF do not want that to be done, there is the potential to use the schema validation and block requests that only change the `modified_comment` field

inventory_management_system_api/repositories/catalogue_item.py
MatteoGuarnaccia5: Regarding the modified_comment field - I put the name in creation as you have the name obviously as it is being created. A property update may be on other fields other than the name. So you cannot put it in the modified comment field.

Deletion I think can have similar - `Property deleted`. It would make sense as it shows up as an `update` event in history. 

For spares, it does get included in history currently, I can modify the pipeline to ignore it so that we omit it from history, or I think it is better to have a simple 'spares updated' modified comment and modified by be 'IMS automation' perhaps? To signify it is an internal mechanism

VKTB:
> If CLF do not want that to be done, there is the potential to use the schema validation and block requests that only change the modified_comment field

Could you please make a note so that you can discuss it with them when you discuss your user stories? Thanks

joelvdavies:
> The `uv run ./scripts/dev_cli.py db-restore mock_data.dump` command from the README for populating the database fails when I ran it with the changes on this branch.

The restore works now, but some of the data doesnt work with the models. The usage statuses fail due to not having a modified by field? I think https://github.com/ral-facilities/inventory-management-system-api/blob/add-last-modified-by-field-%23710/scripts/setup.mongodb needs updating and the mock data regnerated again.

inventory_management_system_api/repositories/catalogue_item.py
joelvdavies: I think that makes sense - If we were to back construct an item at a given point in time, we would need the number of spares for it to be accurate, so best to have it in the history I think. Thanks.

joelvdavies (commented):
After fixing the above issue, its working well for me. I also checked changing the spares definition, the modified by for all items end up changing with "IMS Automation", but I think that makes sense.

joelvdavies (changes requested):
~As I said above, the data needs regenerating too - its broken.~ Sorry, as you said you had updated it its just that its broken still.

joelvdavies (approved):
Working well for me now.

VKTB (approved):

PR #758, opened 2026-04-22
Migrate from `pip` to `uv` #757

### LINKED ISSUE #757: Migrate to uv
Migrate from pip to uv, see guide [here](https://docs.astral.sh/uv/guides/migration/pip-to-project/).

Ensure:
- [x] The dependencies install
- [x] The project installs
- [x] The service runs
- [x] The migration script runs

joelvdavies:
## Description

There is useful information here https://docs.astral.sh/uv/guides/migration/pip-to-project/. But the full steps I followed are found below:
- Add .python-version file
- Convert pyproject.toml & requirements.txt
  - Dependency groups instead of optional dependencies, don’t pin them, just set minimum versions. Replacing `project[group_name]` with `{ include-group="group_name" }` as appropriate
  - Add build-system, and other uv specific sections
  - Keep a record of the dependencies
  - Add requirements with `uv add -r requirements.txt` - this adds all subdependencies to the pyproject.toml even though we only want the direct dependencies – revert the pyproject change by pasting back in the recorded direct dependencies
  - `uv sync`
  - Check environment matches requirements.txt
    - `grep -E '^[A-Za-z0-9_.-]+==' requirements.txt  | tr '[:upper:]' '[:lower:]' | sort > old.txt`
    - `uv export --no-dev --format=requirements-txt --no-hashes | tr '[:upper:]' '[:lower:]' | sort > new.txt`
    - `diff -u old.txt new.txt`
    - Some differences are fine e.g. platform specific things, (Colorama appeared which is windows specific, some things still moved around due to using – instead of _, but versions of existing packages should be identical otherwise)
- Update `Dockerfile`
- Update `dependabot.yml` (Ensure to exclude uv-build as it can cause an issue see https://github.com/ral-facilities/ims-job-scheduler-api/pull/41)

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #757


VKTB:
> * Add requirements with `uv add -r requirements.txt` - this adds all subdependencies – remove them manually again by pasting back in the recorded direct dependencies

What do you mean by "remove them manually again by pasting back in the recorded direct dependencies"?

A note to say that we need to check the IMS System Administrator CLI works.

joelvdavies:
> > * Add requirements with `uv add -r requirements.txt` - this adds all subdependencies – remove them manually again by pasting back in the recorded direct dependencies
> 
> What do you mean by "remove them manually again by pasting back in the recorded direct dependencies"?

It adds all the subdependencies to the pyproject.toml, when we only want it to contain the direct dependencies, I will update to make it clearer.

> A note to say that we need to check the IMS System Administrator CLI works.

Yes sorry, I put the steps as those that would apply everywhere, I did also check that worked, it required adding the extra ENV in the Dockerfile to ensure the script ended up on the path, otherwise a sysadmin would have to also add uv run to the command too.

joelvdavies:
For future reference I just found a better way to migrate

- Run `uvx migrate-to-uv` (Also works for poetry)
- Add a .python-version file
- Update the pyproject.toml
  - Use dependency groups instead of optional dependencies, don’t pin them, just set minimum versions. Replacing `project[group_name]` with `{ include-group="group_name" }` as appropriate
  - Add any other other uv specific sections that appear missing compared to ims-job-scheduler-api
  - Check environment matches requirements.txt (First restore it so can be compared)
    - `grep -E '^[A-Za-z0-9_.-]+==' requirements.txt  | tr '[:upper:]' '[:lower:]' | sort > old.txt`
    - `uv export --no-dev --format=requirements-txt --no-hashes | tr '[:upper:]' '[:lower:]' | sort > new.txt`
    - `diff -u old.txt new.txt`
    - Some differences are fine e.g. platform specific things, (Colorama appeared which is windows specific, some things still moved around due to using – instead of _, but versions of existing packages should be identical otherwise)
- Update the `Dockerfile`
- Update the `dependabot.yml` (Ensure to exclude uv-build as it can cause an issue see https://github.com/ral-facilities/ims-job-scheduler-api/pull/41)
- Update README (Ensure mentions uv as a prerequisite and instances of `python` in commands are replaced by `uv run`)

MatteoGuarnaccia5 (dismissed):
This is working for me!

.github/dependabot.yml
```diff
@@ -50,14 +50,19 @@ updates:
...
-      pip:
+      uv:
         patterns:
           - "*"
+        # Dependabot can check the lockfile, but cannot update these detected in pyproject.toml. As updates are grouped
+        # this causes all updates within the group to be skipped.
+        # E.g. `Skipping change to group uv in directory /: Previous version was not provided for: 'uv-build'`
+        exclude-patterns:
+          - "uv-build"
+          - "inventory-management-system-api"
```
VKTB: Missing newline
```suggestion
          - "inventory-management-system-api"

```

.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
VKTB: Missing newline
```suggestion
3.13

```

VKTB (changes requested):
The `python ./scripts/generate_mock_data.py` command is now failing for me with
```
Traceback (most recent call last):
  File "/inventory-management-system-api/./scripts/generate_mock_data.py", line 9, in <module>
    from faker import Faker
ModuleNotFoundError: No module named 'faker'
```

VKTB (approved):

PR #760, opened 2026-04-24
Catalogue item validation for spreadsheets #759

### LINKED ISSUE #759: Add/Improve validation logic for catalogue items for spreadsheets
For the spreadsheet import functionality we need to be able to get a list of parsable errors from a set of catalogue item spreadsheet data. This task is to work on the actual verification logic that can be used for this. Input should be of catalogue item json data, output should be a list of errors e.g. Pydantic errors.

joelvdavies:
## Description

This PR adds a `validate` function to `CatalogueItemService` taking in catalogue item data (equivalent to a post request), and then produces a list of all the validation warning/errors it can find. The key difference between this and existing validation logic is that this logic compiles a list of all errors rather than failing fast with errors intended to be raised at the router level. The collected errors are all defined in pydantic's format using custom error types to describe which part of the data contains the each issue.

The only warnings raised are from validating a catalogue item that has a duplicate name as one that already exists. This is not as sophisticated as previous logic as its not using indexing nor does it use a code, its merely a warning and not meant to prevent anything but rather warn against it.

## Testing instructions

I added a `/catalogue-item/validate` endpoint for testing, or I can send a test script that uses the service directly.
E.g. Using the mock data
```json
{
    "manufacturer_id": "",
    "name": "Output Couplers",
    "cost_gbp": 10,
    "is_obsolete": false,
    "days_to_replace": "20d",
    "days_to_rework": 10,
    "catalogue_category_id": "69ca7cb64004213685cc98c0",
    "properties": [
        {"id": "69ca7cb64004213685cc98be", "value": "fsdf"},
        {"id": "69ca7cb64004213685cc98bf", "value": "20"}
    ]
}
```
Returns
```json
{
  "warnings": [
    {
      "type": "duplicate_record",
      "location": ["name"],
      "message": "Duplicate record found with the same name 'Output Couplers'",
      "input": "Output Couplers"
    }
  ],
  "errors": [
    {
      "type": "float_parsing",
      "location": ["days_to_replace"],
      "message": "Input should be a valid number, unable to parse string as a number",
      "input": "20d"
    },
    {
      "type": "missing_record",
      "location": ["manufacturer_id"],
      "message": "No manufacturer found with ID ''",
      "input": ""
    },
    {
      "type": "invalid_property_type",
      "location": ["properties", 0, "value"],
      "message": "Invalid value for property with ID '69ca7cb64004213685cc98be'. Expected one of < 0.5% (400 - 700 nm, Average), < 0.3% (450 nm, 700 nm, per Surface, Average), < 1% (per Surface, Average).",
      "input": "fsdf"
    },
    {
      "type": "invalid_property_type",
      "location": ["properties", 1, "value"],
      "message": "Invalid value type for property with ID '69ca7cb64004213685cc98bf'. Expected type: number.",
      "input": "20"
    }
  ]
}
```

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #759


inventory_management_system_api/routers/v1/catalogue_item.py
```diff
@@ -78,6 +79,20 @@ def create_catalogue_item(
...
+@router.post(
+    path="/validate",
+    summary="Validate a catalogue item",
+    response_description="Any errors found in the data",
+)
+def validate_catalogue_item(
+    catalogue_item: Annotated[dict[str, Any], Body(description="Catalogue item data to validate")],
+    catalogue_item_service: CatalogueItemServiceDep,
+) -> ValidationResponseSchema:
+    logger.info("Validating a catalogue item")
```
VKTB, line 91: I have shared some suggestions below that may make it a bit clearer what the endpoint is about. The method docstring is fine to be shown in the Swagger docs in this case.

```suggestion
    summary="Validate a catalogue item for creation",
    response_description="Any errors found in the data",
)
def validate_catalogue_item(
    catalogue_item: Annotated[dict[str, Any], Body(description="Catalogue item data to validate")],
    catalogue_item_service: CatalogueItemServiceDep,
) -> ValidationResponseSchema:
    """
    Validate whether the provided catalogue item data meets the requirements for creating a new catalogue item via the
    catalogue item creation endpoint (`/v1/catalogue-items` `POST`). This includes schema validation, field constraints,
    and business rules, without creating or modifying any resources.
    """
    logger.info("Validating a catalogue item for creation")
```

inventory_management_system_api/routers/v1/catalogue_item.py
```diff
@@ -78,6 +79,20 @@ def create_catalogue_item(
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=message) from exc
 
 
+@router.post(
+    path="/validate",
+    summary="Validate a catalogue item",
+    response_description="Any errors found in the data",
+)
+def validate_catalogue_item(
+    catalogue_item: Annotated[dict[str, Any], Body(description="Catalogue item data to validate")],
```
VKTB, line 88: This is how the example value for the `catalogue_item` body shows up in the Swagger docs, not very useful.

Image

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -256,3 +260,133 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
             ObjectStorageAPIClient.delete_images(catalogue_item_id, access_token)
 
         self._catalogue_item_repository.delete(catalogue_item_id)
+
+    # pylint:disable=too-many-statements
+    def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSchema:
+        """
+        Performs validation of catalogue item creation data returning any errors.
```
VKTB, line 267: Can be expanded based on suggestion I provided above.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -256,3 +260,133 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
             ObjectStorageAPIClient.delete_images(catalogue_item_id, access_token)
 
         self._catalogue_item_repository.delete(catalogue_item_id)
+
+    # pylint:disable=too-many-statements
+    def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSchema:
```
VKTB, line 265: I wonder if it is better to name it `validate_create`? Without reading the doc string, it is not obvious what validation it does, especially given we have other methods like delete, update etc.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -256,3 +260,133 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
...
+            if not catalogue_category:
+                errors.append(
+                    utils.create_custom_validation_error_details(
+                        type="missing_record",
+                        message=f"No catalogue category found with ID '{catalogue_category_id}'",
+                        location=("catalogue_category_id",),
+                        input=catalogue_category_id,
+                    )
+                )
+            # If defined, ensure the category can take new catalogue items (i.e. is a leaf category)
```
VKTB, line 299: Do you mean "if exists" instead of "if defined"?

inventory_management_system_api/services/utils.py
VKTB, line 1: I think this module is becoming difficult to follow/understand especially for someone new to the codebase. I will speak to you to explain what exactly I mean.

inventory_management_system_api/services/utils.py
```diff
@@ -241,3 +398,21 @@ def _merge_non_mandatory_properties(
         elif not defined_property["mandatory"]:
             properties[defined_property_id] = {"id": defined_property_id, "value": None}
     return properties
+
+
+def create_custom_validation_error_details(
+    type: str,  # pylint:disable=redefined-builtin
```
VKTB, line 404: I get this in Pycharm `Shadows built-in name 'type'`.

inventory_management_system_api/services/utils.py
```diff
@@ -241,3 +398,21 @@ def _merge_non_mandatory_properties(
...
+    message: str,
+    location: tuple,
+    input: Any,  # pylint:disable=redefined-builtin
```
VKTB, line 407: I get this in Pycharm `Shadows built-in name 'input'`.

inventory_management_system_api/services/utils.py
```diff
@@ -241,3 +398,21 @@ def _merge_non_mandatory_properties(
...
+    """
+    Creates a Pydantic ValidationError with a custom message and returns its details.
+
+    :param type: String type identifier of the custom error.
+    :param message: Message to display in the error.
+    :param location: Location of the error expressed as a tuple.
+    :param input: Input that led to the error.
+    :return: Created ValidationError.
+    """
+    return InitErrorDetails(type=PydanticCustomError(type, message), loc=location, input=input)
```
VKTB, line 418: I get this for `type` and `message` in Pycharm `Expected type 'LiteralString', got 'str' instead`.

inventory_management_system_api/services/utils.py
joelvdavies: Tricky one to resolve - I can rename them e.g `error_type`/`error_input`?, but they are currently made to match pydantic which purposely uses these names ignoring this, hence why I disabled the pylint warnings.

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Yes I agree I think it is best to rename it.

inventory_management_system_api/routers/v1/catalogue_item.py
joelvdavies: I will make it a very basic example, unfortunately there is no nice way to get it to auto use the schema model. It cant use it directly without performing validation when I dont want it to, and attempts to use example/openapi_extra with it haven't got it to work.

inventory_management_system_api/routers/v1/catalogue_item.py
joelvdavies: Addressed on https://github.com/ral-facilities/inventory-management-system-api/pull/772 to avoid conflicts after switching to bulk.

inventory_management_system_api/routers/v1/catalogue_item.py
joelvdavies: Addressed on https://github.com/ral-facilities/inventory-management-system-api/pull/772 to avoid conflicts after switching to bulk.

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Addressed on https://github.com/ral-facilities/inventory-management-system-api/pull/772 to avoid conflicts after switching to bulk.

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Addressed on https://github.com/ral-facilities/inventory-management-system-api/pull/772 to avoid conflicts after switching to bulk.

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Also ended up fixing on https://github.com/ral-facilities/inventory-management-system-api/pull/772.

inventory_management_system_api/services/utils.py
joelvdavies: I have gone with using `error_` on https://github.com/ral-facilities/inventory-management-system-api/pull/772.

inventory_management_system_api/services/utils.py
joelvdavies: Hopefully fixed now on https://github.com/ral-facilities/inventory-management-system-api/pull/772.

inventory_management_system_api/services/utils.py
joelvdavies: As discussed, have managed to simplify the logic so it avoids duplication and means there are now only _validate methods with an `process_error` function. Tests required minimal changes. See https://github.com/ral-facilities/inventory-management-system-api/pull/772. Also thought similar could be used for the validation there as its similar to the create but havent changed as its not identical and would break the ID handling as I intentionally catch them to allow it to fall back to a not found error and not fail immediately.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -256,3 +260,133 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
...
+        # This records any errors from basic schema validation including the properties
+        try:
+            CatalogueItemPostSchema(**catalogue_item_data)
+        except ValidationError as exc:
+            errors.extend(exc.errors())
+
+        # Check the catalogue category exists (if defined)
+        catalogue_category = None
+        if "catalogue_category_id" in catalogue_item_data:
+            catalogue_category_id = catalogue_item_data["catalogue_category_id"]
```
MatteoGuarnaccia5, line 284:
```suggestion
        catalogue_category = catalogue_item_data.get("catalogue_category_id")
        if catalogue_category_id:
```

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -256,3 +260,133 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
...
+
+        # If defined, check obsolete replacement item exists
+        if "obsolete_replacement_catalogue_item_id" in catalogue_item_data:
+            obsolete_replacement_catalogue_item_id = catalogue_item_data["obsolete_replacement_catalogue_item_id"]
+            if obsolete_replacement_catalogue_item_id is not None:
+                obsolete_replacement_catalogue_item = None
+                try:
+                    obsolete_replacement_catalogue_item = self._catalogue_item_repository.get(
+                        obsolete_replacement_catalogue_item_id
+                    )
```
MatteoGuarnaccia5, line 318:
```suggestion
        # If defined, check obsolete replacement item exists
        obsolete_replacement_catalogue_item_id = catalogue_item_data.get("obsolete_replacement_catalogue_item_id")
            if obsolete_replacement_catalogue_item_id:
                obsolete_replacement_catalogue_item = None
                try:
                    obsolete_replacement_catalogue_item = self._catalogue_item_repository.get(
                        obsolete_replacement_catalogue_item_id
                    )
```

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -256,3 +260,133 @@ def delete(self, catalogue_item_id: str, access_token: Optional[str] = None) ->
...
+                        utils.create_custom_validation_error_details(
+                            type="missing_record",
+                            message=f"No catalogue item found with ID '{obsolete_replacement_catalogue_item_id}'",
+                            location=("obsolete_replacement_catalogue_item_id",),
+                            input=obsolete_replacement_catalogue_item_id,
+                        )
+                    )
+
+        # Check the manufacturer exists (if defined)
+        if "manufacturer_id" in catalogue_item_data:
```
MatteoGuarnaccia5, line 333: Same logic change as above, bit picky but I think is more readable and is fewer lines

inventory_management_system_api/schemas/validation.py
```diff
@@ -0,0 +1,27 @@
...
+
+from typing import Any
+
+from pydantic import BaseModel, Field
+
+
+class ValidationErrorSchema(BaseModel):
+    """Schema model for a validation error."""
+
+    type: str = Field(description="Type of error")
```
MatteoGuarnaccia5, line 13: For maintainability of the codebase, should this be a literal, so when developing you can only use one of the predefined errors, otherwise there is a chance someone could misspell, use a slightly different name, and you end up with non-conformed types in the response.

MatteoGuarnaccia5 (changes requested):

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Thanks, didnt think of that 😄. Like viktors comments above will address on https://github.com/ral-facilities/inventory-management-system-api/pull/772 to avoid conflicts.

inventory_management_system_api/services/catalogue_item.py
joelvdavies: Yeah I prefer it, thanks.

inventory_management_system_api/schemas/validation.py
joelvdavies: I agree it would be nice, unfortunately its not really possible to do this properly, we dont have a complete list of them as some are pydantic specific (see a list here https://pydantic.dev/docs/validation/latest/errors/validation_errors) some are custom from us. Best we can do is add a bunch of constants with different error types for the custom ones.

joelvdavies: Have done that on https://github.com/ral-facilities/inventory-management-system-api/pull/772.

inventory_management_system_api/services/catalogue_item.py
VKTB: As `catalogue_category` is also used outside of this conditional, applying the suggested change would mean that `catalogue_category` could be referenced before assignment later on when validation of the properties is performed. I will check the changes on #772 to see if that is the case.

joelvdavies: Thanks yes it did, adapted to fix.

inventory_management_system_api/repositories/catalogue_item.py
```diff
@@ -278,3 +278,19 @@ def update_number_of_spares(
             {"$set": {"number_of_spares": number_of_spares}},
             session=session,
         )
+
+    def is_duplicate_name(
+        self,
+        name: str,
+        session: Optional[ClientSession] = None,
+    ) -> bool:
```
VKTB: This should fit one one line btw
```suggestion
    def is_duplicate_name(self, name: str, session: Optional[ClientSession] = None) -> bool:
```

test/unit/services/test_utils.py
VKTB, line 1: Looks like when `name` was removed from `PropertyPostSchema` a while back, the tests in this file were not updated to reflect this. Do you mind removing `name` in the places where it is passed to create `PropertyPostSchema` instances? Does `mock_data.py` needs to be updated too?

VKTB (changes requested):

inventory_management_system_api/repositories/catalogue_item.py
joelvdavies: Thanks yes, ruff and black keep conflicting, if I forget to change my vscode default back it does this and never fixes it unless I remove the last column 😄

joelvdavies: Fixed here.

test/unit/services/test_utils.py
joelvdavies: Sure, I will have a look on https://github.com/ral-facilities/inventory-management-system-api/pull/772.

joelvdavies: Yes both name and unit_id have been removed from the schema, so I have removed those. On the mock data, we dont use ids for them as they change which is why the name field is being used instead then they are matched in the e2e tests, as mentioned in the comments e.g.
```
basic catalogue item data as would be required
for a `CatalogueItemPostSchema` but with mandatory IDs missing and any `id`'s
replaced by the `name` value in its properties as the IDs will be added automatically.
```
There arent any unit_ids there either, so I don't believe there is anything in the mock_data to change.

VKTB (approved):

MatteoGuarnaccia5 (approved):

PR #766, opened 2026-05-12
Release/v4.0.1

VKTB:
## Description
Makes a release for v4.0.1. Containing the latest package updates.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build


VKTB:
Because the project version goes in the uv lockfile, it means that it needs to be updated with `uv lock` after manually updating the version in the `pyproject.toml` file. That's why the unit test CI job was failing as the Docker stage was erroring.

joelvdavies (approved):

PR #772, opened 2026-05-20
Add bulk creation of catalogue items #764

### LINKED ISSUE #764: Add bulk catalogue item creation endpoint
The spreadsheet import requires a place to send a group of catalogue items (validated via #759), to create all at once in a transaction so they can be guarenteed to all succeed/fail together.

joelvdavies:
## Description

Adds a POST `/catalogue-items/bulk` endpoint taking a list of catalogue items and creating them all at once in a transaction. This endpoint fails fast resulting in it being harder to determine which item specifically caused an error e.g. if a manufactuerer is not found, it will fail but it wont say on which item. This is made up for by the validate endpoint that all validation errors it finds.

## Other changes
- Changes `/catalogue-items/validate` to `/catalogue-items/bulk-validate`. This is the same as before, but now takes a list so more than one catalogue item can be validated in a single request.

## Notes

- The implementation is done as the simplest implementation possible reusing existing validation logic that is already part of the creation request. The most optimal solution would instead to process the group of catalouge items and then do an insert many, having done all fail fast validation before hand, but that is a much larger change and seemed unecessary at this point for a one off new feature. (It would also effect history)
- E2E tests only test some of the creation logic e.g. There are various ways a `InvalidPropertyTypeError` can occur, but its only testing one as it just needs to check the catch/raise logic in the router. The standard create tests are otherwise testing the rest of the logic as the existing `create` methods are being used in the `bulk_create` anyway.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #764

### DIFF
--- a/inventory_management_system_api/routers/v1/catalogue_item.py
+++ b/inventory_management_system_api/routers/v1/catalogue_item.py
@@ -89,7 +89,10 @@ def bulk_create_catalogue_item(
 ) -> list[CatalogueItemSchema]:
     logger.info("Bulk creating catalogue items")
     try:
-        return catalogue_item_service.bulk_create(catalogue_items)
+        return [
+            CatalogueItemSchema(**catalogue_item.model_dump())
+            for catalogue_item in catalogue_item_service.bulk_create(catalogue_items)
+        ]
     except (InvalidPropertyTypeError, MissingMandatoryProperty) as exc:
         logger.exception(str(exc))
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=str(exc)) from exc
--- a/inventory_management_system_api/routers/v1/catalogue_item.py
+++ b/inventory_management_system_api/routers/v1/catalogue_item.py
@@ -86,24 +86,15 @@ def create_catalogue_item(
 )
 def bulk_create_catalogue_item(
     catalogue_items: list[CatalogueItemPostSchema], catalogue_item_service: CatalogueItemServiceDep
-):
+) -> list[CatalogueItemSchema]:
     logger.info("Bulk creating catalogue items")
     try:
-        catalogue_item_service.bulk_create(catalogue_items)
+        return catalogue_item_service.bulk_create(catalogue_items)
     except (InvalidPropertyTypeError, MissingMandatoryProperty) as exc:
         logger.exception(str(exc))
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=str(exc)) from exc
     except (MissingRecordError, InvalidObjectIdError) as exc:
-        if "catalogue category" in str(exc).lower():
-            message = "A specified catalogue category does not exist"
-            logger.exception(message)
-            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=message) from exc
-        if "manufacturer" in str(exc).lower():
-            message = "A specified manufacturer does not exist"
-            logger.exception(message)
-            raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=message) from exc
-
-        message = "A specified replacement catalogue item does not exist"
+        message = "A specified entity does not exist"
         logger.exception(message)
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_CONTENT, detail=message) from exc
     except NonLeafCatalogueCategoryError as exc:
@@ -122,7 +113,7 @@ def bulk_validate_create_catalogue_item(
         list[dict[str, Any]],
         Body(
             description="List of catalogue items data to validate",
-            example=[{"name": "Catalogue Item 1"}, {"name": "Catalogue Item 2"}],
+            examples=[[{"name": "Catalogue Item 1"}, {"name": "Catalogue Item 2"}]],
         ),
     ],
     catalogue_item_service: CatalogueItemServiceDep,
--- a/inventory_management_system_api/routers/v1/catalogue_item.py
+++ b/inventory_management_system_api/routers/v1/catalogue_item.py
@@ -113,17 +113,31 @@ def bulk_create_catalogue_item(
 
 
 @router.post(
-    path="/bulk-validate",
-    summary="Bulk validate a catalogue items",
+    path="/bulk-validate-create",
+    summary="Bulk validate catalogue items for creation",
     response_description="Any errors found in the data",
 )
-def validate_catalogue_item(
-    catalogue_items: Annotated[list[dict[str, Any]], Body(description="List of catalogue items data to validate")],
+def bulk_validate_create_catalogue_item(
+    catalogue_items: Annotated[
+        list[dict[str, Any]],
+        Body(
+            description="List of catalogue items data to validate",
+            example=[{"name": "Catalogue Item 1"}, {"name": "Catalogue Item 2"}],
+        ),
+    ],
     catalogue_item_service: CatalogueItemServiceDep,
 ) -> BulkValidationResultSchema:
-    logger.info("Bulk validating catalogue items")
+    """
+    Validate a list of catalogue item data to determine whether it is suitable for creating them as catalogue items
+    either via the single `POST /v1/catalogue-items` or the bulk `POST /v1/catalogue-items/bulk` endpoints.
 
-    return catalogue_item_service.bulk_validate(catalogue_items)
+    This includes the same validation of schema, field constraints and business rules, as these creation endpoints
+    without creating or modifying any resources. This also collects and returns any validation errors instead of failing
+    fast.
+    """
+    logger.info("Bulk validating catalogue items for creation")
+
+    return catalogue_item_service.bulk_validate_create(catalogue_items)
 
 
 @router.get(path="", summary="Get catalogue items", response_description="List of catalogue items")
--- a/inventory_management_system_api/routers/v1/catalogue_item.py
+++ b/inventory_management_system_api/routers/v1/catalogue_item.py
@@ -11,10 +11,21 @@ service.
 import logging
 from typing import Annotated, Any, List, Optional
 
-from fastapi import APIRouter, Body, Depends, HTTPException, Path, Query, Request, status
+from fastapi import (
+    APIRouter,
+    Body,
+    Depends,
+    HTTPException,
+    Path,
+    Query,
+    Request,
+    status,
+)
 
 from inventory_management_system_api.core.config import config
-from inventory_management_system_api.core.consts import HTTP_500_INTERNAL_SERVER_ERROR_DETAIL
+from inventory_management_system_api.core.consts import (
+    HTTP_500_INTERNAL_SERVER_ERROR_DETAIL,
+)
 from inventory_management_system_api.core.exceptions import (
     ChildElementsExistError,
     InvalidActionError,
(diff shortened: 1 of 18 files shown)


inventory_management_system_api/routers/v1/catalogue_item.py
```diff
@@ -80,17 +80,55 @@ def create_catalogue_item(
...
+    either via the single `POST /v1/catalogue-items` or the bulk `POST /v1/catalogue-items/bulk` endpoints.
+
+    This includes the same validation of schema, field constraints and business rules, as these creation endpoints
+    without creating or modifying any resources. This also collects and returns any validation errors instead of failing
+    fast.
+    """
+    logger.info("Bulk validating catalogue items for creation")
 
-    return catalogue_item_service.validate(catalogue_item)
+    return catalogue_item_service.bulk_validate_create(catalogue_items)
```
VKTB, line 134: I get `Expected type 'dict[str, Any]', got 'list[dict[str, Any]]' instead` warning for `catalogue_items`.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -279,59 +310,58 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
...
-            if not catalogue_category:
+            if catalogue_category is None:
                 errors.append(
                     utils.create_custom_validation_error_details(
-                        type="missing_record",
-                        message=f"No catalogue category found with ID '{catalogue_category_id}'",
-                        location=("catalogue_category_id",),
-                        input=catalogue_category_id,
+                        error_type=ERROR_TYPE_MISSING_RECORD,
+                        error_message=f"No catalogue category found with ID '{catalogue_category_id}'",
```
VKTB, line 325: I get `Expected type 'LiteralString', got 'str' instead` warnings for both `error_type` and `error_message`.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -279,59 +310,58 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
...
-            # If defined, ensure the category can take new catalogue items (i.e. is a leaf category)
+            # If exists, ensure the category can take new catalogue items (i.e. is a leaf category)
             elif not catalogue_category.is_leaf:
                 errors.append(
                     utils.create_custom_validation_error_details(
-                        type="non_leaf_catalogue_category",
-                        message="Cannot add catalogue item to a non-leaf catalogue category",
-                        location=("catalogue_category_id",),
-                        input=catalogue_category_id,
+                        error_type=ERROR_TYPE_NON_LEAF_CATALOGUE_CATEGORY,
```
VKTB, line 334: I get `Expected type 'LiteralString', got 'str' instead` warning for `error_type`.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -279,59 +310,58 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
...
+                    obsolete_replacement_catalogue_item_id
+                )
+            except InvalidObjectIdError:
+                # Ignore invalid object ID, treat as missing
+                pass
+            if obsolete_replacement_catalogue_item is None:
+                errors.append(
+                    utils.create_custom_validation_error_details(
+                        error_type=ERROR_TYPE_MISSING_RECORD,
+                        error_message=f"No catalogue item found with ID '{obsolete_replacement_catalogue_item_id}'",
```
VKTB, line 356: I get `Expected type 'LiteralString', got 'str' instead` warnings for both `error_type` and `error_message`.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -376,7 +406,7 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
 
             # Perform validation of the properties - can only be done assuming a valid catalogue category has been found
             if catalogue_category:
-                utils.validate_properties(catalogue_category.properties, property_schemas, errors)
+                utils.process_properties(catalogue_category.properties, property_schemas, errors)
```
VKTB: Following on from my https://github.com/ral-facilities/inventory-management-system-api/pull/760#discussion_r3312381361, that there is a possibility that `catalogue_category` has not been assigned.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -341,10 +371,10 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
             if not manufacturer:
                 errors.append(
                     utils.create_custom_validation_error_details(
-                        type="missing_record",
-                        message=f"No manufacturer found with ID '{manufacturer_id}'",
-                        location=("manufacturer_id",),
-                        input=manufacturer_id,
+                        error_type=ERROR_TYPE_MISSING_RECORD,
+                        error_message=f"No manufacturer found with ID '{manufacturer_id}'",
```
VKTB, line 375: I get `Expected type 'LiteralString', got 'str' instead` warnings for both `error_type` and `error_message`.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -353,10 +383,10 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
             if self._catalogue_item_repository.is_duplicate_name(catalogue_item_data["name"]):
                 warnings.append(
                     utils.create_custom_validation_error_details(
-                        type="duplicate_record",
-                        message=f"Duplicate record found with the same name '{catalogue_item_data["name"]}'",
-                        location=("name",),
-                        input=catalogue_item_data["name"],
+                        error_type=ERROR_TYPE_DUPLICATE_RECORD,
+                        error_message=f"Duplicate record found with the same name '{catalogue_item_data["name"]}'",
```
VKTB, line 387: I get `Expected type 'LiteralString', got 'str' instead` warnings for both `error_type` and `error_message`.

inventory_management_system_api/services/catalogue_item.py
```diff
@@ -387,6 +417,20 @@ def validate(self, catalogue_item_data: dict[str, Any]) -> ValidationResponseSch
...
+    def bulk_validate_create(self, catalogue_items_data: dict[str, Any]) -> BulkValidationResultSchema:
+        """
+        Performs validation of bulk catalogue item creation data returning any errors.
+
+        :param catalogue_items_data: Catalogue items data to verify.
+        :return: Schema containing the validation warnings/errors that been found within the data.
+        """
+        return BulkValidationResultSchema(
+            results=[
+                self._validate_create(index, catalogue_item_data)
```
VKTB, line 434: I get `Expected type 'dict[str, Any]', got 'str' instead` warning for `catalogue_item_data`.

inventory_management_system_api/services/utils.py
```diff
@@ -73,49 +107,24 @@ def process_properties(
...
-    :param supplied_properties: The list of supplied property objects.
-    :param errors: List to append any found errors to (its mutable).
-    """
     # Convert properties to dictionaries for easier lookups
     defined_properties_dict = _create_properties_dict(defined_properties)
     supplied_properties_dict = _create_properties_dict(supplied_properties)
 
     # Some mandatory properties may not have been supplied
-    _validate_missing_mandatory_properties(defined_properties_dict, supplied_properties_dict, errors)
+    _validate_missing_mandatory_properties(defined_properties_dict, supplied_properties_dict, process_error)
```
VKTB, line 122: I get `Expected type '(str, str, tuple, Any) -> None', got '(error_type: LiteralString, error_message: LiteralString, _error_location: tuple, error_input: Any) -> NoReturn | (error_type: LiteralString, error_message: LiteralString, error_location: tuple, error_input: Any) -> None' instead` warning for `process_error`.

inventory_management_system_api/services/utils.py
```diff
@@ -124,7 +133,9 @@ def validate_properties(
     # Supplied properties do not have units as we can't trust they would be correct
     _add_property_units(defined_properties_dict, supplied_properties_dict)
     # The values of the supplied properties may not be of the expected types
-    _validate_property_values(defined_properties_dict, supplied_properties_dict, errors)
+    _validate_property_values(defined_properties_dict, supplied_properties_dict, process_error)
```
VKTB, line 133: I get the same warning as above for `process_error`.

inventory_management_system_api/services/utils.py
```diff
@@ -248,26 +214,22 @@ def _validate_property_value(
...
-                    "invalid_property_type",
-                    message=f"Mandatory property with ID '{supplied_property_id}' cannot be None.",
-                    location=("properties", index, "value"),
-                    input=supplied_property_value,
-                )
+            process_error(
+                error_type=ERROR_TYPE_INVALID_PROPERTY_TYPE,
+                error_message=f"Mandatory property with ID '{supplied_property_id}' cannot be None.",
+                error_location=("properties", index, "value"),
+                error_input=supplied_property_value,
```
VKTB, line 218: I get `Unexpected argument` warning for all these here and in all the places in this module where `process_error` is called. Tho that's probably because Pycharm can't work out the stuff that is going on with `process_error` and its type.

VKTB (commented):
First pass

inventory_management_system_api/services/utils.py
joelvdavies: Hopefully fixed now.

inventory_management_system_api/services/utils.py
joelvdavies: The type was wrong, so hopefully fixed now?

VKTB: This still shows up but a note to say that we discussed it and decided to keep it as it is. If we don't specify the argument names then it would be fine. Alternative is to use `Protocol` from `typing`.

inventory_management_system_api/routers/v1/catalogue_item.py
```diff
@@ -80,17 +80,55 @@ def create_catalogue_item(
...
+    path="/bulk",
+    summary="Bulk create new catalogue items",
+    status_code=status.HTTP_201_CREATED,
+)
+def bulk_create_catalogue_item(
+    catalogue_items: list[CatalogueItemPostSchema], catalogue_item_service: CatalogueItemServiceDep
+) -> list[CatalogueItemSchema]:
+    logger.info("Bulk creating catalogue items")
+    try:
+        return catalogue_item_service.bulk_create(catalogue_items)
```
VKTB: For the return here I get `Expected type 'list[CatalogueItemSchema]', got 'list[CatalogueItemOut]' instead`.

test/e2e/test_catalogue_item.py
```diff
@@ -662,6 +662,234 @@ def test_create_with_invalid_manufacturer_id(self):
...
+
+    def test_bulk_create(self):
+        """Test bulk creating catalogue items."""
+
+        self.post_catalogue_item_prerequisites_no_properties()
+        self.post_bulk_catalogue_items(
+            [CATALOGUE_ITEM_DATA_REQUIRED_VALUES_ONLY, CATALOGUE_ITEM_DATA_NOT_OBSOLETE_NO_PROPERTIES]
+        )
+
+        self.check_post_bulk_catalogue_items_success(
```
VKTB, line 769: I get `Expected type 'dict', got 'list[Any]' instead` warning.

VKTB (changes requested):
I am testing the `/v1/catalogue-items/bulk-validate-create` endpoint as per instructions in https://github.com/ral-facilities/inventory-management-system-api/pull/760 but the `missing_record` error for `manufacturer_id` is not in the response. Also, I noticed 200 is being used when errors are returned, shouldn't this be used only if validation succeeds?

joelvdavies:
> I am testing the `/v1/catalogue-items/bulk-validate-create` endpoint as per instructions in #760 but the `missing_record` error for `manufacturer_id` is in the response. Also, I noticed 200 is being used when errors are returned, shouldn't this be used only if validation succeeds?

Thanks, yes it should have been present but wasnt. Fixed now. On the status code, I had a look for recommendations on what to use and it seems 200 is the correct response. As a HTTP request, the validation is successful as long as it can perform the validation successfully, not whether there are validation errors found or not, as unlike the create for example, it doesnt rely on the data being good in order to perform the request which in this case is to validate the data, not use it to do something.

VKTB (changes requested):
I am testing the `/v1/catalogue-items/bulk` `POST` endpoint, sending a request with the below data in the body which fails with `{"detail": "Missing mandatory property with ID '69ca7cb64004213685cc98be'"}` as expected.
```
[
    {
        "manufacturer_id": "69ca7cb64004213685cc989b",
        "name": "Output Couplerss",
        "cost_gbp": 10,
        "is_obsolete": false,
        "days_to_replace": 20,
        "days_to_rework": 10,
        "catalogue_category_id": "69ca7cb64004213685cc98c0"
    }
]
```

However, when I send a request with the same body to the `/v1/catalogue-items/bulk-validate-create` `POST` endpoint, it returns no errors implying the validation was successful.
```
{
  "results": [
    {
      "index": 0,
      "warnings": [],
      "errors": []
    }
  ]
}
```

I am not sure how much these changes have manually been tested but I think we need to do more testing. Also not sure why the `test_validate_create_with_missing_mandatory_properties` e2e is passing.

joelvdavies:
> /v1/catalogue-items/bulk-validate-create

Ah thanks I see the issue. Its because "properties" isnt being supplied at all - I will fix it and add a test case. Most of my testing has been through the spreadsheet, so didnt pick this up as they are always defined.

I have done some more manual testing and havent managed to find any other issues. I realised I wasnt properly testing the property schema validation so added that to the existng invalid schema tests too.

VKTB:
> I am not sure how much these changes have manually been tested but I think we need to do more testing.

I couldn't think of any more edge cases.

VKTB (dismissed):

VKTB (approved):

MatteoGuarnaccia5 (approved):

PR #778, opened 2026-05-26
Add limit to bulk creation and validation of catalogue items #777

### LINKED ISSUE #777: Add max catalogue item limit to bulk validation and creation
To prevent large DOS requests from being performed. In theory non-authenticated requests should be rejected before any schema level processing occurrs anyway so the issue is minimal (the larger is that of starlette potentially loading this data before executing anything but that can only be solved by a proxy level limit instead), and it appears that the recommendation is to have multiple guards rather than just one.

Using a before_validator is likely the best that can be done here for this case.

joelvdavies:
## Description

Adds new `BULK__MAX_CATALOGUE_ITEMS` config value providing an upper limit to the number of catalogue items that can be validated in the new bulk creation/validation endpoints. Uses a before validator to ensure no checking of the schema itself e.g. conversion to the pydantic schema occurrs before checking the size to dismiss requests that are too large immediately.

Defaults to 1000 which takes ~2 seconds for me, so leaves room to increase if necessary.

Also appliesa few minor changes from the setup fo the ingest api, namely removing the recommendation of using `--dev` with `uv sync` as its unecessary, and removign the TZ env variable from the dockerfile as was only required for one of job scheduler's underlying libraries and was accidentally copied here.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking

Closes #777


inventory_management_system_api/routers/v1/catalogue_item.py
```diff
@@ -79,13 +80,30 @@ def create_catalogue_item(
...
+
+
 @router.post(
     path="/bulk",
     summary="Bulk create new catalogue items",
     status_code=status.HTTP_201_CREATED,
 )
 def bulk_create_catalogue_item(
-    catalogue_items: list[CatalogueItemPostSchema], catalogue_item_service: CatalogueItemServiceDep
+    catalogue_items: Annotated[list[CatalogueItemPostSchema], BeforeValidator(check_bulk_max)],
```
VKTB: As discussed limiting the length of the list by utilising `Field` (like below) seems to be doing the length validation before it does the schema validation.


```suggestion
    catalogue_items: Annotated[list[CatalogueItemPostSchema], Field(max_length=config.bulk.max_catalogue_items)],
```

VKTB (dismissed):

MatteoGuarnaccia5 (dismissed):

VKTB (approved):

MatteoGuarnaccia5 (approved):

PR #807, opened 2026-07-06
Replace invalid action error with NonLeafCatalogueCategoryError

### LINKED ISSUE #381: Unify NonLeafCatalogueCategoryError and InvalidActionError in CC repsponses
See note on https://github.com/ral-facilities/inventory-management-system-api/pull/379. We currently raise `NonLeafCatalogueCategoryError` as tested in `test_create_with_non_leaf_catalogue_category`. But for the catalogue category property endpoints we instead use an `InvalidActionError`. These should be unified.

Reillyhewitson:
## Description
I changed a logic error where it produced the wrong type of error, it should have produced a NonLeafCatalogueCategoryError but instead it outputted and InvalidActionError to the correct error which is NonLeafCatalogueCategoryError.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

Detail testing
- [x] Get the list of Catalogue Categories in the swagger docs
- [x] Choose a non leaf Catalogue Category (where `is_leaf` is false) 
- [x] Copy the ID of this Catalogue Categoory
- [x] Go to the properties endpoint
- [x] Add the non-leaf ID and remove the Allowed values field (if you do not want to submit any data to test)
- [x] Submit and the error should be `422: Cannot add a property to a non-leaf catalogue category`  

## Agile board tracking

closes #381

### DIFF
--- a/inventory_management_system_api/services/catalogue_category_property.py
+++ b/inventory_management_system_api/services/catalogue_category_property.py
@@ -8,7 +8,7 @@ from typing import Annotated, Optional
 from fastapi import Depends
 
 from inventory_management_system_api.core.database import start_session_transaction
-from inventory_management_system_api.core.exceptions import InvalidActionError, MissingRecordError
+from inventory_management_system_api.core.exceptions import InvalidActionError, MissingRecordError, NonLeafCatalogueCategoryError
 from inventory_management_system_api.models.catalogue_category import (
     AllowedValues,
     CatalogueCategoryPropertyIn,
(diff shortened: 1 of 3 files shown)


inventory_management_system_api/services/catalogue_category_property.py
```diff
@@ -67,7 +71,7 @@ def create(
         :param catalogue_category_property: Property to add (with additional info on how to perform the migration if
                                         necessary)
         :raises InvalidActionError: If attempting to add a mandatory property without a default_value being specified
-                                    or if the catalogue category is not a leaf
+        :raises NonLeafCatalogueCategoryError: If the catalogue category is not a leaf
```
joelvdavies:
```suggestion
        :raises NonLeafCatalogueCategoryError: If the catalogue category is not a leaf category.
```
Very picky suggestion, just to match the same comment found in `catalogue_category_property.py`.

joelvdavies (commented):
Changes look good to me, nicely done!

Just one very minor comment. I would also recommend editing the PR description to have "Closes" on the last line instead of "connects to" as then it will auto close the issue when its merged.

joelvdavies (approved):
LGTM!

